PR #128, opened 2024-11-14
feat: Implement `getTokenInfo` (foundry solution)

### LINKED ISSUE #108: [Core] Implement `getTokenInfo` on HTS to retrieve the treasury account and token info
HTS allows the user to set up a treasury account which acts as the receiver when newly minted tokens are created. In order to support mint (and burn) operations[^1], we need to have access to the treasury account. The treasury account[^2] is retrievable through the `getTokenInfo(address)`[^3] as specified in [HIP-514](https://hips.hedera.com/hip/hip-514).

Implement the `getTokenInfo(address)` as part of `HtsSystemContract`. Provide the necessary slots to store the corresponding information per token, _i.e._, `TokenInfo`[^4]

```solidity
    struct TokenInfo {
        /// Basic properties of a Hedera Token
        HederaToken token;

        /// The number of tokens (fungible) or serials (non-fungible) of the token
        int64 totalSupply;

        /// Specifies whether the token is deleted or not
        bool deleted;

        /// Specifies whether the token kyc was defaulted with KycNotApplicable (true) or Revoked (false)
        bool defaultKycStatus;

        /// Specifies whether the token is currently paused or not
        bool pauseStatus;

        /// The fixed fees collected when transferring the token
        FixedFee[] fixedFees;

        /// The fractional fees collected when transferring the token
        FractionalFee[] fractionalFees;

        /// The royalty fees collected when transferring the token
        RoyaltyFee[] royaltyFees;

        /// The ID of the network ledger
        string ledgerId;
    }
```

[^1]: https://github.com/hashgraph/hedera-forking/issues/78
[^2]: [IHederaTokenService.sol#L148 `treasury`](https://github.com/hashgraph/hedera-smart-contracts/blob/6d4bb129089770f96c3ca48dd043d68295558a2d/contracts/system-contracts/hedera-token-service/IHederaTokenService.sol#L148)
[^3]: [IHederaTokenService.sol#L668-L674 `getTokenInfo`](https://github.com/hashgraph/hedera-smart-contracts/blob/6d4bb129089770f96c3ca48dd043d68295558a2d/contracts/system-contracts/hedera-token-service/IHederaTokenService.sol#L668-L674)
[^4]: [IHederaTokenService.sol#L173-L200 `TokenInfo`](https://github.com/hashgraph/hedera-smart-contracts/blob/6d4bb129089770f96c3ca48dd043d68295558a2d/contracts/system-contracts/hedera-token-service/IHederaTokenService.sol#L173-L200)

victor-yanev:
**Description**:

**Related issue(s)**:

Fixes #108

**Notes for reviewer**:

**Checklist**


jeromy-cannon (dismissed):
only approving .gitignore

src/MirrorNode.sol
```diff
@@ -2,8 +2,10 @@
 pragma solidity ^0.8.0;
 
 import {Vm} from "forge-std/Vm.sol";
+import {IMirrorNode} from "./IMirrorNode.sol";
+import {storeAddress, storeBool} from "./StrStore.sol";
```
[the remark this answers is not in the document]
victor-yanev: Nope, removed them

src/HtsSystemContractJson.sol
```diff
@@ -32,6 +34,13 @@ contract HtsSystemContractJson is HtsSystemContract {
         return uint32(bytes4(keccak256(abi.encodePacked(account))));
     }
 
+    function getTokenInfo(address token) htsCall external override returns (int64 responseCode, TokenInfo memory tokenInfo) {
+        require(token != address(0), "getTokenInfo: invalid token");
+        string memory json = mirrorNode().fetchTokenData(token);
+        tokenInfo = _getTokenInfo(json);
+        responseCode = 22; // SUCCESS
```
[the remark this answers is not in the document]
victor-yanev: Not really, I guess if the mirror node is not available?

src/StrStore.sol
```diff
@@ -33,3 +39,23 @@ function storeString(address target, uint256 slot, string memory strvalue) {
...
+
+function storeBool(address target, uint256 slot, bool value) {
+    bytes32 boolData = value ? bytes32(uint256(1)) : bytes32(uint256(0));
+    vm.store(target, bytes32(slot), boolData);
+}
+
+function storeAddress(address target, uint256 slot, address value) {
+    bytes32 addressData = bytes32(uint256(uint160(value)));
+    vm.store(target, bytes32(slot), addressData);
+}
```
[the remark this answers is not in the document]
victor-yanev: Yes, in `_initTokenData()` of `HtsSystemContractJson.sol`

src/MirrorNode.sol
```diff
@@ -2,8 +2,9 @@
 pragma solidity ^0.8.0;
 
 import {Vm} from "forge-std/Vm.sol";
+import {IMirrorNodeResponses} from "./IMirrorNodeResponses.sol";
 
-abstract contract MirrorNode {
+abstract contract MirrorNode is IMirrorNodeResponses {
```
acuarica, line 7: Is this needed?

src/MirrorNode.sol
```diff
@@ -51,13 +52,36 @@ abstract contract MirrorNode {
...
+        try this.fetchAccount(accountId) returns (string memory json) {
+            if (vm.keyExistsJson(json, ".evm_address")) {
+                return vm.parseJsonAddress(json, ".evm_address");
+            }
+        } catch {
+            // Do nothing
+        }
+
+        // ignore the first 4 characters ("0.0.") to get the account number string
+        require(bytes(accountId).length > 4, "Invalid account ID, needs to be in the format '0.0.<account_number>'");
```
acuarica, line 71: Either we should perform a stricter validation or just change the message to be more precise

```suggestion
        require(bytes(accountId).length > 4, "Invalid account ID");
```

src/MirrorNode.sol
```diff
@@ -51,13 +52,36 @@ abstract contract MirrorNode {
         return 0;
     }
 
+    function getAccountAddress(string memory accountId) external returns (address) {
+        if (bytes(accountId).length == 0
+            || keccak256(bytes(accountId)) == keccak256(bytes("null"))
+            || keccak256(abi.encodePacked(accountId)) == keccak256(abi.encodePacked(bytes32(0)))) {
```
acuarica, line 58: question: all these different variants come from the Mirror Node?

test/HTS.t.sol
```diff
@@ -54,4 +54,67 @@ contract HTSTest is Test, TestSetup {
...
+        assertEq(tokenInfo.token.memo, "USDC HBAR");
+        assertEq(tokenInfo.token.tokenSupplyType, false);
+        assertEq(tokenInfo.token.maxSupply, 0);
+        assertEq(tokenInfo.token.freezeDefault, false);
+        assertEq(tokenInfo.token.tokenKeys.length, 7);
+        // AdminKey
+        assertEq(tokenInfo.token.tokenKeys[0].keyType, 0x1);
+        assertEq(tokenInfo.token.tokenKeys[0].key.inheritAccountKey, false);
+        assertEq(tokenInfo.token.tokenKeys[0].key.contractId, address(0));
+        assertEq(tokenInfo.token.tokenKeys[0].key.ed25519, vm.parseBytes("5db29fb3f19f8618cc4689cf13e78a935621845d67547719faf49f65d5c367cc"));
```
acuarica, line 78: nit: can we use `hex` literal instead?

test/HTS.t.sol
```diff
@@ -54,4 +54,67 @@ contract HTSTest is Test, TestSetup {
...
+        assertEq(tokenInfo.fixedFees.length, 0);
+        assertEq(tokenInfo.fractionalFees.length, 0);
+        assertEq(tokenInfo.royaltyFees.length, 0);
+        assertEq(tokenInfo.ledgerId, testMode == TestMode.FFI ? "0x01" : "0x00");
+    }
+
+    function test_HTS_getTokenInfo_should_revert_for_invalid_token() external {
+        address token = address(0);
+        vm.expectRevert(bytes("getTokenInfo: invalid token"));
+        HtsSystemContract(HTS_ADDRESS).getTokenInfo(token);
```
acuarica, line 112: Is this the behavior of HTS? Maybe here the `responseCode` is used.

We could also include another case with `makeAddr` in addition to `address(0)`.

test/HTS.t.sol
```diff
@@ -54,4 +54,67 @@ contract HTSTest is Test, TestSetup {
...
+    }
+
+    function test_HTS_getTokenInfo_should_revert_when_delegatecall() external {
+        vm.expectRevert(bytes("htsCall: delegated call"));
+        (bool revertsAsExpected, ) = HTS_ADDRESS.delegatecall(abi.encodeWithSelector(HtsSystemContract.getTokenInfo.selector, address(this)));
+        assertTrue(revertsAsExpected, "expectRevert: call did not revert");
+    }
 }
```
acuarica, line 120: I would also add another test that should revert when calling `getTokenInfo` through a token, `IHederaTokenService(token).getTokenInfo`. This is to ensure the `__redirectForToken` validation works as expected.

src/StrStore.sol
```diff
@@ -7,7 +7,17 @@ Vm constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
...
+    storeBytes(target, slot, value);
+}
+
+function storeBytes(address target, uint256 slot, bytes memory value) {
     uint256 length = value.length;
+
+    if (length == 0) {
+        vm.store(target, bytes32(slot), bytes32(0));
+        return;
+    }
```
acuarica, line 19: why is this special case needed?

src/HtsSystemContractJson.sol
```diff
@@ -66,6 +67,479 @@ contract HtsSystemContractJson is HtsSystemContract {
...
         assembly { slot := initialized.slot }
         vm.store(address(this), slot, bytes32(uint256(1)));
+
+        _initTokenInfo(json);
+    }
+
+    function _initTokenInfo(string memory json) internal {
+        TokenInfo memory tokenInfo = _getTokenInfo(json);
+
+        bytes32 slot;
```
acuarica, line 77: nit: wouldn't be simple if this is `uint256`?

src/HtsSystemContractJson.sol
```diff
@@ -66,6 +67,479 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+            revert("getTokenInfo: Token pause status is required");
+        }
+
+        return tokenInfo;
+    }
+
+    function _getHederaToken(string memory json) private returns (HederaToken memory token) {
+        token.tokenKeys = _getTokenKeys(json);
+
+        try vm.parseJsonString(json, ".name") returns (string memory name) {
```
acuarica, line 276: I would remove the `try/catch/revert` altogether.

src/HtsSystemContractJson.sol
```diff
@@ -66,6 +67,479 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+        // https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html
+        p = uint256(slot);
+        itemSlot = uint256(keccak256(abi.encodePacked(p)));
+        storeUint(address(this), p, tokenInfo.fixedFees.length);
+        for (uint256 i = 0; i < tokenInfo.fixedFees.length; i++) {
+            storeInt(address(this), itemSlot, tokenInfo.fixedFees[i].amount);
+            itemSlot += 1;
+
+            storeAddress(address(this), itemSlot, tokenInfo.fixedFees[i].tokenId);
+            offset = 20;
```
acuarica, line 166: I think we shouldn't support `offset`s. This adds unnecessary complexity to both Foundry's and Hardhat's solution. I guess it's just easier to add _gaps_ into the struct.

acuarica (changes requested):
Nice work! left some comments

src/MirrorNode.sol
victor-yanev: Nope, `parseJsonString` resolves to `"null"` when the field in the json has a `null` value, other checks are just safety checks to confirm that the given address is valid before calling mirror node

src/StrStore.sol
victor-yanev: For some reason `storeBytes` was behaving strangely with empty bytes, this addition fixed the problem

acuarica: We could add some unit tests for `storeBytes` in `StrStore.t.sol` to detect what was/is the problem. This feels like a workaround for (maybe) a problem somewhere else. Given the low-level byte manipulation we are doing, we need to ensure this does not hide any problem elsewhere.

victor-yanev: Hmmm, tbh I don't remember exactly why I added this...

Looking at it right now, it should behave exactly like the `if (length <= 31)` block:
```solidity
        // this would be the same as: slotValue = bytes32(new bytes(0)) | bytes32(0) = bytes32(0)
        bytes32 slotValue = bytes32(value) | bytes32(length * 2);
        // thus, this would actually do: storeBytes32(target, slot, bytes32(0)), which is the same
        storeBytes32(target, slot, slotValue);
        return;
```

victor-yanev: I removed the `if (length == 0)` block, cause it seemed to make no difference when running the unit tests

test/lib/TestSetup.sol
```diff
@@ -50,18 +52,21 @@ abstract contract TestSetup {
...
     function setUpMockStorageForNonFork() internal {
         if (HTS_ADDRESS.code.length == 0) {
             console.log("HTS code length is 0, non-fork test, code and data provided locally");
-            MirrorNodeMock mirrorNode = new MirrorNodeMock();
-            mirrorNode.deployHIP719Proxy(USDC, "USDC");
-            mirrorNode.deployHIP719Proxy(MFCT, "MFCT");
+            MirrorNodeMock mirrorNodeMock = new MirrorNodeMock();
+            mirrorNodeMock.deployHIP719Proxy(USDC, "USDC");
+            mirrorNodeMock.deployHIP719Proxy(MFCT, "MFCT");
+            mirrorNode = mirrorNodeMock;
```
acuarica, line 63: What is the intent of this change? What does it add?

src/HtsSystemContract.sol
```diff
@@ -38,6 +40,23 @@ contract HtsSystemContract is IERC20Events {
...
+    /**
+     * Query token info
+     * @param token - The token address to check
+     * @return responseCode - the response code for the status of the request. SUCCESS is `22`.
+     * @return tokenInfo - token info for `token`
+     */
+    function getTokenInfo(address token) htsCall external view returns (int64 responseCode, TokenInfo memory tokenInfo) {
+        require(token != address(0), "getTokenInfo: invalid token");
+
+        (bool success, bytes memory data) = token.staticcall(
```
acuarica, line 52: can we replace this `staticcall` with `IHederaTokenService(token).getTokenInfo(token)`? We would need to remove `view`, but this is how it's declared in `IHederaTokenService` so it should be fine.

test/lib/TestSetup.sol
victor-yanev:
```solidity
    function test_HTS_getTokenInfo_should_revert_when_call_to_mirror_node_fails() external {
         address token = MFCT;
         vm.mockCallRevert(address(mirrorNode), abi.encode(mirrorNode.fetchTokenData.selector), abi.encode("mirror node error"));
         vm.expectRevert(bytes("getTokenInfo: failed to get token info"));
         HtsSystemContract(HTS_ADDRESS).getTokenInfo(token);
     }
```

acuarica (approved):

PR #132, opened 2024-11-18
feat: cache Mirror Node responses

### LINKED ISSUE #115: [Hardhat] Reuse responses from Mirror Node's `getToken` endpoint
Currently, the Hardhat plugin sends a request to the Mirror Node for each `eth_getStorageAt` call. This is necessary to provide data like balances and allowances. However, when retrieving token metadata like `name`, `symbol` or `decimals`, the response from the Mirror Node[^1] can be reused. This is might not seem a big issue at the moment. But when `getTokenInfo`[^2] gets implemented, this will trigger several `eth_getStorageAt` calls.

Reuse the response from the `getToken` endpoint, to avoid making duplicated requests to the Mirror Node.

[^1]: https://testnet.mirrornode.hedera.com/api/v1/docs/#/tokens/getToken
[^2]: https://github.com/hashgraph/hedera-forking/issues/108

acuarica:
**Description**:

This PR enhances the JSON-RPC Forwarder to cache (in memory) the responses from the Mirror Node. This is particularly useful in two scenarios

1. When retrieving `timestamp`s the `blockNumber` usually doesn't change, which means the `timestamp` remains the same.
2. When retrieving Token metadata fields, _i.e._, `name`, `symbol`, `decimals` and `totalSupply`. All these fields request information from the same Mirror Node endpoint, `/tokens/{tokenId}`. Moreover, a `string` request (`name` and `symbol`) may trigger multiple `eth_getStorageAt` calls. Given the `blockNumber` is used, we can cache the response of these endpoint.

To that end, `getHtsStorageAt` and Mirror Node client methods have been modified to receive `blockNumber` as parameters.

Additional changes

- Renamed `hardhatAddresses` to `localAccount`. This is to make the Forwarder more general, because it's not Hardhat specific.
- Renamed `projectTestConfig` module in Hardhat tests to make clear its purpose.
- Enhance JSON-RPC Mock to be able to run as a Worker. In this case, it uses `postMessage` to notify the parent that the process is ready for connections.
- Include unit tests for `getHIP719Code` to ensure code is loaded properly.
- Update `README`'s sequence diagram to show new architecture.

**Related issue(s)**:

Fixes #115.

**Notes for reviewer**:

> [!NOTE]
> Both `logger` and `reqId` were removed from `getHtsStorageAt`. These two parameters were initially included because `getHtsStorageAt` was meant to be part of the JSON-RPC Relay. Given we changed the architecture, and now it's part of a local Hardhat plugin, there is no need for these to parameters.

> [!NOTE]
> The `anvil.test.js` starts and terminates processes. We need to ensure that process termination is correct. I tried to use https://github.com/wevm/prool to instantiate `anvil`, but it requires Node 22 (they use a new function [`Promise.withResolvers`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers#browser_compatibility)). I didn't want to constraint Node version just because of this.


rbarkerSL (approved):
Approval applies to `README.md`

victor-yanev (approved):

PR #142, opened 2024-11-27
feat: implement `getTokenInfo` for Hardhat

### LINKED ISSUE #141: [Hardhat] Implement `getTokenInfo` state on HTS for Hardhat
As a follow up of https://github.com/hashgraph/hedera-forking/issues/108, we need to implement the `getTokenInfo` state for Hardhat. This implies translating from slot numbers in `eth_getStorageAt` to Mirror Node fields, and apply any conversion necessary.

acuarica:
**Description**:

This PR implements `getTokenInfo` for Hardhat. It extends the `getHtsStorageAt` to interpret Solidity structs and arrays together with support for more types. Also supports extracting keys and custom fees (except royalty fees which are only applicable to NFTs).

It builds a "`SlotMap`" for a given `token`. This is a mapping from slots to primitive values. It uses the Solidity [storage layout](https://docs.soliditylang.org/en/latest/internals/layout_in_storage.html) definition to build the slot map.

Additionally, this PR includes a script to create tokens with custom fees to aid testing f5eef27a0d055adea1b29df646e69bb5c4b96e77, a144b29151ba905c3102c8860d2f950126800e3b, 55f688c81695100dd6d8b3cf233ede15c9821ace. The `launch-token` submodule was removed in favor of this new script 7292022d5a5b2ada5fb634484d1ec1b3898c73ee.

**Related issue(s)**:

Fixes #141.

**Notes for reviewer**:

When testing fractional fees, two issues appeared in the Foundry solution. One was it was unable to `abi.decode`. I had to include `all_collectors_are_exempt` in `src/IMirrorNodeResponses.sol`. The other issue was that field values were shifted. I guess there was a missing `++` in `src/HtsSystemContractJson.sol`. See this changes in 40893c8.

**Checklist**


@hts-forking/src/slotmap.js
```diff
@@ -0,0 +1,158 @@
...
+                      ),
+                  ]
+                : [hexstr, len * 2, []];
+
+        return [`${header.padEnd(64 - 2, '0')}${lenByte.toString(16).padStart(2, '0')}`, ...chunks];
+    },
+    t_uint8: val => [toIntHex256(val)],
+    t_uint256: val => [toIntHex256(val)],
+    t_int256: str => [toIntHex256(str ?? 0)],
+    t_address: str => [toIntHex256(str?.replace('0.0.', '') ?? 0)],
```
victor-yanev, line 73: 🤔 , wouldn't it be better if we called mirror node to get the real EVM address of the account instead of relying on the long-zero address?

@hts-forking/src/slotmap.js
```diff
@@ -0,0 +1,158 @@
...
+        assert('_type' in key && 'key' in key);
+        if (key._type === 'ED25519')
+            return { key_type, ed25519: key.key, _e_c_d_s_a_secp256k1: '' };
+        return { key_type, ed25519: '', _e_c_d_s_a_secp256k1: key.key };
+    });
+
+    const customFees = /**@type {Record<string, unknown>}*/ (token['custom_fees']);
+    token['fixed_fees'] = customFees['fixed_fees'] ?? [];
+    token['fractional_fees'] = customFees['fractional_fees'] ?? [];
+    token['royalty_fees'] = customFees['royalty_fees'] ?? [];
```
victor-yanev, line 109: I don't think this would work, because the actual `TokenInfo` struct has very different fields for the custom fees from those which are in the mirror node response

@hts-forking/src/slotmap.js
```diff
@@ -0,0 +1,158 @@
...
+
+/**
+ * @param {Record<string, unknown>} token
+ * @returns {Map<bigint, {value: string, path: string}>}
+ */
+function slotMapOf(token) {
+    token['default_kyc_status'] = false;
+    token['treasury'] = token['treasury_account_id'];
+    token['ledger_id'] = '0x00';
+    token['second'] = `${token['expiry_timestamp']}`;
```
victor-yanev, line 85: Isn't this supposed to be part of the `expiry` struct? Why are we mapping it directly to a `second` field? Also, what about the auto-renewal account?

@hts-forking/src/slotmap.js
```diff
@@ -0,0 +1,158 @@
...
+    t_int256: str => [toIntHex256(str ?? 0)],
+    t_address: str => [toIntHex256(str?.replace('0.0.', '') ?? 0)],
+    t_bool: value => [toIntHex256(value ? 1 : 0)],
+};
+
+/**
+ * @param {Record<string, unknown>} token
+ * @returns {Map<bigint, {value: string, path: string}>}
+ */
+function slotMapOf(token) {
```
victor-yanev, line 81: This method seems really hacky and tries to hit multiple rabbits with a single bullet. The `flatten` method also seems to have multiple purposes and is really hard to read and follow through.

Wouldn't it be better if we just constructed a new object that would represent the `TokenInfo` struct and mapped the mirror node response to it? 

For example, one way to do it can be by defining a new class that will represent the `TokenInfo` struct:
- in its constructor (or through a static method - e.g., `fromMirrorNodeToken`) it can be created from the mirror node object - something similar to what is done in `_initTokenInfo` in the foundry solution where the `TokenInfo` struct is being populated from the mirror node JSON response.

Then in order to fill the slots we would have to:
1. Construct an instance of the `TokenInfo` object, by populating it from the token response from mirror node
2. Iterate through the fields of the `TokenInfo` object and use the storage layout to figure out the slots where each field should be residing

This seems way easier to understand and also would be very similar to what is done in the `HtsSystemContractJson._initTokenInfo`  method

victor-yanev (changes requested):

@hts-forking/src/slotmap.js
acuarica: It is, but these `struct`s are flattened, so it uses the last part of field path, _i.e._, `.second`.

For the same reasons, the auto-renewal fields didn't need any extra care.

@hts-forking/src/slotmap.js
acuarica: Yes, true that. Going to fix it.

@hts-forking/src/slotmap.js
acuarica: Done.

@hts-forking/src/slotmap.js
acuarica: I think it would be better to tackle this in another PR. The reason being, our tests don't cover custom fees. This implies creating custom tokens with the desired fees, which in itself seems out of scope.

@hts-forking/src/slotmap.js
acuarica: Mmm, so _1._ and _2._ is essentially what `slotMapOf` is doing. The first bit corresponds loosely to _1._, but just amending the `token` so it's compatible with the Solidity definition. The second part is doing _2._, but driven by the storage layout instead.

`getHtsStorageAt` is driven by slot number (not fields), so to me it's more clear to follow the storage layout first.

@hts-forking/src/slotmap.js
victor-yanev: Sure, sounds good. 

Btw, what I did to test things out for the custom fees (while doing the foundry solution) was to mock some data inside the `json` files for `USDC` and `MFCT` tokens and compare the results from `getTokenInfo` with the data

@hts-forking/src/slotmap.js
victor-yanev: Alright, makes sense. Let's add a small comment to clarify this.

@hts-forking/src/slotmap.js
victor-yanev: Okay, can we create a new object, instead of amending the token object? Also, in the new object we can directly write all fields explicitly in camelCase (as they are in the solidity struct), instead of snake case (this would skip the conversion logic). 

In this way it would be more explicit and clearer what fields exist in the mapped object and what we are doing, amending the token object is kind of ambiguous because we are only adding the missing fields and requires us to believe that the actual response will contain the rest of the fields, instead of catching any potential problems (missing fields, etc) on time.

@hts-forking/src/slotmap.js
acuarica: Do you have any examples? Did you create the data manually? I would like to use the same test data for both Foundry and Hardhat.

@hts-forking/src/slotmap.js
acuarica: I wanted to avoid re-creating the whole object, just wanted to adapt what is needed in order to retrieve the field values. This way we reuse the storage layout (if we create a new object we are repeating ourselves by duplicating the storage layout).

@hts-forking/src/slotmap.js
victor-yanev: Yes, I found some real examples from random tokens on testnet and copied the custom fees from there

victor-yanev (approved):

@hts-forking/src/slotmap.js
```diff
@@ -0,0 +1,175 @@
...
+const { keccak256 } = require('ethers');
+
+const { toIntHex256, toSnakeCase } = require('./utils');
+const {
+    storageLayout: { storage, types },
+} = require('../resources/HtsSystemContract.json');
+
+/** @typedef {string | ((mirrorNode: Pick<import('.').IMirrorNodeClient, 'getAccount'>, blockNumber: number) => Promise<string>)} Value */
+
+class SlotMap {
```
Nana-EC, line 29: Is this class logic called out in a doc or current README or in process to document.
I think it's really powerful.
At the least you should update the PR description a bit more

Nana-EC (approved):
Nice work.
SlotMap is great stuff, can' you update the PR description to summarize logic there but I'm also looking forward to in depth description of it elsewhere

@hts-forking/src/slotmap.js
acuarica: Ok, at the end I managed to include some custom fee examples, so no need for a follow-up ticket for this.

victor-yanev (approved):
LGTM

PR #143, opened 2024-11-27
feat: [Foundry] Fetch HIP-719 remote state for fungible tokens

### LINKED ISSUE #98: [Foundry] Fetch HIP-719 remote state for fungible tokens
As a follow up of https://github.com/hashgraph/hedera-forking/issues/82, fetch remote state related to `associate`, `dissociate` and `isAssociated` methods for Foundry contract.

The remote state should come from `accounts/{accountId}/tokens?token.id={tokenId}`[^1] endpoint.

> [!NOTE]
> Currently, the Mirror Node _only_ allows the user to fetch associations without specifying a "block number"/timestamp. See https://github.com/hashgraph/hedera-forking/issues/96 for more details.

[^1]: https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getTokensByAccountId

victor-yanev:
**Description**:

**Related issue(s)**:

Fixes #98

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/scripts/curl
+++ b/scripts/curl
@@ -84,6 +84,21 @@ for (const [re, fn] of /** @type {const} */([
             return undefined;
         }
     }],
+    [/^accounts\/((0x[0-9a-fA-F]{40})|(0\.0\.\d+))\/tokens/, idOrAliasOrEvmAddress => {
+        assert(typeof idOrAliasOrEvmAddress === 'string');
+        const tokenId = searchParams.get('token.id');
+        assert(tokenId !== null);
+
+        const token = tokens[tokenId];
+        if (token === undefined)
+            return { tokens: [], links: { next: null } };
+
+        try {
+            return require(`../@hts-forking/test/data/${token.symbol}/getTokenRelationship_${idOrAliasOrEvmAddress.toLowerCase()}.json`);
+        } catch {
+            return undefined;
+        }
+    }],
 ])) {
     const match = endpoint.match(re);
     if (match !== null) {
--- a/src/HtsSystemContract.sol
+++ b/src/HtsSystemContract.sol
@@ -213,14 +213,17 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
             emit Approval(owner, spender, amount);
             return abi.encode(true);
         } else if (selector == IHRC719.associate.selector) {
+            _initTokenRelationships(msg.sender);
             bytes32 slot = _isAssociatedSlot(msg.sender);
             assembly { sstore(slot, true) }
             return abi.encode(true);
         } else if (selector == IHRC719.dissociate.selector) {
+            _initTokenRelationships(msg.sender);
             bytes32 slot = _isAssociatedSlot(msg.sender);
             assembly { sstore(slot, false) }
             return abi.encode(true);
         } else if (selector == IHRC719.isAssociated.selector) {
+            _initTokenRelationships(msg.sender);
             bytes32 slot = _isAssociatedSlot(msg.sender);
             bool res;
             assembly { res := sload(slot) }
@@ -244,6 +247,9 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
     }
 
     function _initTokenData() internal virtual {
+    }
+
+    function _initTokenRelationships(address account) internal virtual {
     }
 
     function _balanceOfSlot(address account) internal virtual returns (bytes32) {
--- a/src/HtsSystemContractJson.sol
+++ b/src/HtsSystemContractJson.sol
@@ -14,6 +14,7 @@ contract HtsSystemContractJson is HtsSystemContract {
     MirrorNode private _mirrorNode;
 
     bool private initialized;
+    bool private relationshipsInitialized;
 
     function setMirrorNodeProvider(MirrorNode mirrorNode_) htsCall external {
         _mirrorNode = mirrorNode_;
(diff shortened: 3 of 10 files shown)


src/HtsSystemContractJson.sol
```diff
@@ -81,6 +82,35 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+            string memory notFoundError = "{\"_status\":{\"messages\":[{\"message\":\"Not found\"}]}}";
+            if (keccak256(bytes(json)) == keccak256(bytes(notFoundError))) {
+                storeBool(address(this), uint256(slot), false);
+            } else {
+                bytes memory tokens = vm.parseJson(json, ".tokens");
+                IMirrorNodeResponses.TokenRelationship[] memory relationships = abi.decode(tokens, (IMirrorNodeResponses.TokenRelationship[]));
+                storeBool(address(this), uint256(slot), relationships.length > 0);
+            }
+        } catch {
+            storeBool(address(this), uint256(slot), false);
```
acuarica, line 102: Wouldn't be better to left it untouched when error?

src/HtsSystemContractJson.sol
```diff
@@ -81,6 +82,35 @@ contract HtsSystemContractJson is HtsSystemContract {
...
 
+    function _initTokenRelationships(address account) internal override {
+        if (relationshipsInitialized[account]) {
+            // Already initialized
+            return;
+        }
+
+        bytes32 slot = _isAssociatedSlot(account);
+        try mirrorNode().fetchTokenRelationshipOfAccount(vm.toString(account), address(this)) returns (string memory json) {
+            string memory notFoundError = "{\"_status\":{\"messages\":[{\"message\":\"Not found\"}]}}";
```
acuarica, line 93: We can use `keyExistsJson` instead of comparing with the error as we did it in other places.

src/HtsSystemContractJson.sol
```diff
@@ -81,6 +82,35 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+            if (keccak256(bytes(json)) == keccak256(bytes(notFoundError))) {
+                storeBool(address(this), uint256(slot), false);
```
acuarica, line 95: also we here we should left it untouched?

src/HtsSystemContract.sol
```diff
@@ -213,14 +213,17 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
             emit Approval(owner, spender, amount);
             return abi.encode(true);
         } else if (selector == IHRC719.associate.selector) {
+            _initTokenRelationships(msg.sender);
```
acuarica, line 216: This should be similar to `_balanceOfSlot`, maybe we can override `_isAssociatedSlot` and the the mechanisms used there.

acuarica (changes requested):
looking good

scripts/curl
```diff
@@ -84,6 +84,21 @@ for (const [re, fn] of /** @type {const} */([
             return undefined;
         }
     }],
+    [/^accounts\/((0x[0-9a-fA-F]{40})|(0\.0\.\d+))\/tokens/, idOrAliasOrEvmAddress => {
```
Nana-EC, line 87: I think we should have tests that ensure these regexs work and catch regressions.
Not an item for this PR as I see this file has the same issue in multiple places.
Could you create a follow up ticket to capture this ask

src/IMirrorNodeResponses.sol
```diff
@@ -39,4 +39,14 @@ interface IMirrorNodeResponses {
         int64 amount;
         string denominating_token_id;
     }
+
+    struct TokenRelationship {
```
Nana-EC, line 43: Q: Does this represent a Token entity or the mapping of ownership by an account holding a token?

Nana-EC (commented):
Looking good, some considerations

scripts/curl
acuarica: This scripts is used for testing (to provide a drop-in replacement for `curl` while running tests to avoid hitting the real Mirror Node). Do we still want to create tests for this?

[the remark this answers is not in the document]
acuarica: I guess same issue as https://github.com/hashgraph/hedera-forking/pull/143#discussion_r1866990174 applies.

src/IMirrorNodeResponses.sol
acuarica: It represents the response of this endpoint https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getTokensByAccountId from the Mirror Node (without the enclosing array because we provide the `token.id` when querying the Mirror Node).

src/IMirrorNodeResponses.sol
```diff
@@ -39,4 +39,14 @@ interface IMirrorNodeResponses {
...
+        bool automatic_association;
+        uint256 balance;
+        string created_timestamp;
+        uint256 decimals;
+        string token_id;
```
acuarica, line 48: to be able to use it with `parseJson`/`decode`, does this need to be sorted alphabetically?

vyanev1: I think it is indifferent to the order, but I will need to confirm

victor-yanev: I copied the same order that we get from the response by the mirror node, I think decode is indifferent to the order, but I will need to confirm.

src/IMirrorNodeResponses.sol
victor-yanev: Yes, this is the `TokenRelationship` response object described in the schema for: https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getTokensByAccountId

```solidity
if (vm.keyExistsJson(json, ".tokens")) {
  bytes memory tokens = vm.parseJson(json, ".tokens");
  IMirrorNodeResponses.TokenRelationship[] memory relationships = abi.decode(tokens, (IMirrorNodeResponses.TokenRelationship[]));
  return relationships.length > 0;
}
```
- If the `".tokens"` field exists:
  - we try to parse the array of token relationships
  - since we have passed a `token.id` param with the token ID, if the array is not empty it means that the given token is associated with the account and we return `true`
- If the `".tokens"` field doesn't exist:
  - most likely Mirror Node returned `{"_status":{"messages":[{"message":"Not found"}]}}`, which means the given account address doesn't exist
  - in this case, we just ignore the error and return a default response of `false`

acuarica (approved):

scripts/curl
victor-yanev: +1, we use this `curl` override script only for the tests

PR #144, opened 2024-12-03
feat: [Hardhat] Fetch HIP-719 remote state for fungible tokens

### LINKED ISSUE #126: [Hardhat] Fetch HIP-719 remote state for fungible tokens
As a follow up of https://github.com/hashgraph/hedera-forking/issues/82, fetch remote state related to `associate`, `dissociate` and `isAssociated` methods for Hardhat plugin.

The remote state should come from `accounts/{accountId}/tokens?token.id={tokenId}`[^1] endpoint.

> [!NOTE]
> Currently, the Mirror Node _only_ allows the user to fetch associations without specifying a "block number"/timestamp. See https://github.com/hashgraph/hedera-forking/issues/96 for more details.

[^1]: https://testnet.mirrornode.hedera.com/api/v1/docs/#/accounts/getTokensByAccountId

victor-yanev:
**Description**:

Adds needed logic for the `isAssociated` method to the `@hts-forking/src/index.js` file.

On match of the `isAssociated` selector, the hardhat relay fetches the state from the `accounts/{accountId}/tokens?token.id={tokenId}` MN endpoint and returns the result of the token relationship.

**Related issue(s)**:

Fixes #126 

**Notes for reviewer**:

**Checklist**


@hts-forking/scripts/json-rpc-mock.js
```diff
@@ -121,6 +121,15 @@ const mirrorNodeClient = {
             noAllowance
         );
     },
+    async getTokenRelationship(idOrAliasOrEvmAddress, tokenId) {
+        this.append('getTokenRelationship', idOrAliasOrEvmAddress, tokenId);
+        const noTokens = { tokens: [] };
```
Nana-EC, line 126: nit: pull `noTokens` out so that it's static and not created on each call.

@hts-forking/test/getHtsStorageAt.test.js
```diff
@@ -38,7 +38,10 @@ describe('::getHtsStorageAt', function () {
 
     /** @type {IMirrorNodeClient} */
     const baseMirrorNodeClient = {
-        getTokenById(_tokenId) {
+        async getTokenById(_tokenId) {
+            throw Error('Not implemented');
```
Nana-EC, line 42: Q: Why are these tests not implemented? Is there a plan to?

@hts-forking/src/index.js
```diff
@@ -168,6 +168,22 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
         );
     }
 
+    // Encoded `address(tokenId).isAssociated()` slot
+    // slot(256) = `isAssociated`selector(32) + padding(192) + accountId(32)
+    if (nrequestedSlot >> 224n === 0x4d8fdd6dn) {
```
Nana-EC, line 173: Is there a test that confirms this mapping for `0x4d8fdd6dn` works and ensures no regression.
I don't see one in the PR but let me know if it already exists.

Nana-EC (commented):
Nice work, some thoughts, also please update the description to be a bit more descriptive

@hts-forking/src/index.js
victor-yanev: Yes, it is tested through the `test_IHRC719_with_real_accounts()` in `IHRC719.t.sol`

@hts-forking/test/getHtsStorageAt.test.js
victor-yanev: Not sure why they are not implemented yet, but it looks like it’s out of scope for this PR

acuarica: > Q: Why are these tests not implemented? Is there a plan to?

These are meant to be `Not implemented`. This `baseMirrorNodeClient` acts as a base object to be used in subsequent tests where you override what it's needed for said test. This is to ensure no unwanted calls to the Mirror Node are made.

@hts-forking/src/index.js
acuarica: Two things here

**1.** I'd add a unit test that covers this case. You can include it in `getHtsStorageAt.test.js`. There you'll find similar cases for balances and allowances in there.
**2.** Also please include a validation right after the `if` to ensure that the padding is zero. Given the association signature, the padding should be the same as `balanceOf`. Otherwise return `null`.

@hts-forking/test/data/MFCT/getTokenRelationship_0.0.2669675.json
```diff
@@ -0,0 +1,16 @@
+{
```
acuarica, line 1: can we reuse the same data files we had for Foundry?

@hts-forking/scripts/json-rpc-mock.js
acuarica: @Nana-EC I'd usually prefer to have definitions in their more specific scope (unless there is another reason to not to). This is to make it clear to see where is being used. If it's in outer scope, I might think it's used in several places, which is not.

acuarica (changes requested):
looking good, left some comments above

@hts-forking/test/getHtsStorageAt.test.js
victor-yanev: So it's more of a stub,

 I suggest we rename it to `mirrorNodeClientStub` or `mockMirrorNodeClient` to make things clearer.

@hts-forking/test/data/MFCT/getTokenRelationship_0.0.2669675.json
victor-yanev: Kinda not - because we are calling it in a different way here - by ID and in foundry we directly pass the address of the accounts

@hts-forking/src/index.js
victor-yanev: 1. Done, added unit tests both for MFCT and USDC tokens with the following test cases:
```javascript
{
    name: 'token relationship is found',
    fn: (/** @type {string} */ idOrAliasOrEvmAddress, /** @type {string} */ _tid) =>
        require(`./data/${symbol}/getTokenRelationship_${idOrAliasOrEvmAddress}`),
},
{
    name: 'token relationship is empty',
    fn: (/** @type {string} */ _accountId, /** @type {string} */ _tid) => ({ tokens: [] }),
},
{
  name: 'token relationship is null',
  fn: (/** @type {string} */ _accountId, /** @type {string} */ _tid) => null,
},
```
2. Changed it to account for padding inside the if itself (similar to how it is done for the `balanceOf` selector:
```javascript
    if (
        nrequestedSlot >> 32n ===
        0x4d8fdd6d_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000n
    ) {
      ...
    }
```

@hts-forking/src/index.js
```diff
@@ -168,6 +168,25 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
...
+    // slot(256) = `isAssociated`selector(32) + padding(192) + accountId(32)
+    if (
+        nrequestedSlot >> 32n ===
+        0x4d8fdd6d_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000_0000n
+    ) {
+        const accountId = `0.0.${parseInt(requestedSlot.slice(-8), 16)}`;
+        const { tokens } = (await mirrorNodeClient.getTokenRelationship(accountId, tokenId)) ?? {
+            tokens: [],
+        };
+        if (tokens.some(t => t.token_id === tokenId)) {
```
acuarica, line 181: Can we replace this with checking `tokens.length`, given the `tokenId` was already provided in the MN query?

@hts-forking/src/index.js
```diff
@@ -168,6 +168,25 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
...
+            return ret(
+                `0x${Number(1).toString(16).padStart(64, '0')}`,
```
acuarica, line 183: nit: can we use `toIntHex256`? 

in `main` there is a new version of this function that accepts `number`.

@hts-forking/test/data/MFCT/getTokenRelationship_0.0.2669675.json
acuarica: Ok, then let's use a symlink instead of duplicating the file (to make it clear both file represent the same content).

Similar to what it's done in the `getTokenInfo` for Hardhat PR

```console
ln -s getAccount_0.0.2669675.json getAccount_0xa3612a87022a4706fc9452c50abd2703ac4fd7d9.json
```

acuarica (approved):

PR #145, opened 2024-12-04
feat: Implement HTS emulation for NFT methods

### LINKED ISSUE #130: [Core] Implement HTS emulation for NFT methods
In addition to ERC20, HTS contract should provide emulation for NFT/ERC721 methods. These methods are specified in [HIP-218](https://hips.hedera.com/hip/hip-218).

victor-yanev:
**Description**:

This PR adds the necessary changes to `__redirectForToken` in order to handle `ERC721`(NFT) methods. 

These methods are specified in [HIP-218](https://hips.hedera.com/hip/hip-218).

**Related issue(s)**:

Fixes #130

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,10 @@ This is necessary because our library relies on [`curl`](https://curl.se/) to ma
 This enables the library to fetch token state in the remote network.
 Given `curl` is an external command, `ffi` needs to be enabled.
 
+> [!NOTE]
+> The Foundry library assumes it is running on a Unix-like operating system.
+> In addition to `curl`, it uses `bash`, `tail`, `sed` and `tr` and to make requests and parse responses.
+
 To activate HTS emulation in your tests, you need to add the following setup code in your test files.
 Import our wrapper function to deploy HTS emulation and enable cheat codes for it.
 
@@ -391,6 +395,23 @@ The following methods can be invoked on the Hedera Token Service contract locate
 | `mintToken(address token, int64 amount, bytes[] metadata)`      | Mints an amount of the token to the defined treasury account   |
 
 <!-- -->
+
+## JS Library
+
+Service and application builders can use the core state emulation library.
+These functions can be embedded in a JSON-RPC implementation to enable System Contracts fork testing without using the Hardhat plugin.
+
+```javascript
+const { getHtsCode, getHtsStorageAt, getHIP719Code } = require('@hashgraph/system-contracts-forking');
+```
+
+- **`getHtsCode`**. Gets the runtime bytecode for the Solidity implementation of the HTS System Contract.
+- **`getHtsStorageAt`**. Allows the user to query the state of an HTS Token as if it were a standard smart contract.
+- **`getHIP719Code`**. Returns the token proxy contract bytecode for the given `address`. Based on the proxy contract defined by [HIP-719](https://hips.hedera.com/hip/hip-719).
+
+See [`index.d.ts`](./src/index.d.ts) for more details.
+
+At its core, the [_Hardhat plugin_](#hardhat-plugin) is supported by these functions.
 
 ## Build
 
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -148,11 +148,15 @@ export const LONG_ZERO_PREFIX: string;
 export function getHIP719Code(address: string): string;
 
 /**
- * Gets the bytecode for the Solidity implementation of the HTS System Contract.
+ * Gets the runtime bytecode for the Solidity implementation of the HTS System Contract.
+ *
+ * This bytecode is meant to be deployed at address `0x167`.
  */
 export function getHtsCode(): string;
 
 /**
+ * Allows the user to query the state of an HTS Token as if it were a standard smart contract.
+ *
  * This function should not throw, provided the `mirrorNodeClient` does not throw either.
  * If the `mirrorNodeClient` throws, _e.g._, due to connection issues,
  * error should be handled by the caller.
@@ -160,15 +164,15 @@ export function getHtsCode(): string;
  * When the token ID corresponding to `address` does not exist,
  * the respective calls on `mirrorNodeClient` should return `null`.
  *
- * The storage mechanism for `balanceOf` and `allowance` use a map between addresses and account IDs.
+ * The storage mechanism for _"dynamic"_ slots, _e.g._, `balanceOf` or `allowance`, use a map between addresses and account IDs.
  * This allow the contract to reduce the space to marshal an account:
  * `32 bits` (or even `64 bits` if longer IDs are needed) using `accountid` (omitting the `shardId` and `realmId`) against `160 bits` using `address`.
  * This mechanism in turn allow us to marshal more than one account used in storage slots, _e.g._, `allowance`.
  *
- * @param address
- * @param slot
- * @param blockNumber
- * @param mirrorNodeClient
+ * @param address can be either a Token address or the HTS address (`0x167`).
+ * @param slot corresponds to the slots defined by the Storage Layout of `HtsSystemContract` or a _"dynamic"_ slot.
+ * @param blockNumber specifies the point in time to fetch the requested `slot` from the Mirror Node.
+ * @param mirrorNodeClient the client used to fetch Token data from the Mirror Node.
  */
 export function getHtsStorageAt(
     address: string,


src/HtsSystemContract.sol
```diff
@@ -212,7 +238,50 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
...
+            uint256 tokenId = uint256(bytes32(msg.data[40:72]));
+            return abi.encode(__ownerOf(tokenId));
+        } else if (selector == IERC721.transferFrom.selector) {
+            require(msg.data.length >= 124, "transferFrom: Not enough calldata");
+            address from = address(bytes20(msg.data[40:60]));
+            address to = address(bytes20(msg.data[72:92]));
+            uint256 tokenId = uint256(bytes32(msg.data[92:124]));
+            _transferNFT(from, to, tokenId);
+            return abi.encode(true);
+        } else if (selector == IERC721.approve.selector) {
```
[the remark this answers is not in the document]
victor-yanev: Good point, I guess we will need to check whether the token is fungible or not and then use the appropriate `redirect` function

src/HtsSystemContract.sol
```diff
@@ -212,7 +238,50 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
...
+        return abi.encode("undefined");
+    }
+
+    function __redirectForERC721(bytes4 selector) internal virtual returns (bytes memory) {
+        if (selector == IERC721.ownerOf.selector) {
+            require(msg.data.length >= 60, "ownerOf: Not enough calldata");
+            _initTokenData();
+            uint256 tokenId = uint256(bytes32(msg.data[40:72]));
+            return abi.encode(__ownerOf(tokenId));
+        } else if (selector == IERC721.transferFrom.selector) {
```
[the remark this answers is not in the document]
victor-yanev: No, currently it's not reachable

src/HtsSystemContract.sol
```diff
@@ -160,7 +161,32 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
      */
     function __redirectForToken() internal virtual returns (bytes memory) {
         bytes4 selector = bytes4(msg.data[24:28]);
+        bytes memory result = abi.encode(false);
 
+        result = __redirectForERC20(selector);
```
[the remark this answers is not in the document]
victor-yanev: I think it's better not to couple them together and have different methods for FTs and NFTs, even though we might have a small amount of duplication (e.g., for `name` and `symbol`)

src/HtsSystemContract.sol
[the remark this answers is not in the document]
victor-yanev: Please, check the changes after newest commit.

src/HtsSystemContract.sol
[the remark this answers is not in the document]
victor-yanev: Please, check the changes after newest commit.

test/ERC721.t.sol
```diff
@@ -0,0 +1,107 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import {HtsSystemContract, HTS_ADDRESS} from "../src/HtsSystemContract.sol";
+import {Test, console} from "forge-std/Test.sol";
```
acuarica, line 5: if not needed, we can remove `console` here.

src/HtsSystemContract.sol
```diff
@@ -1,16 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
+import {console} from "forge-std/Console.sol";
```
acuarica, line 4: we should remove this, right?

src/HtsSystemContract.sol
```diff
@@ -160,33 +163,69 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
      */
     function __redirectForToken() internal virtual returns (bytes memory) {
         bytes4 selector = bytes4(msg.data[24:28]);
+        bytes memory result = abi.encode(false);
+        _initTokenData();
+
+        // Redirect to the appropriate ERC20 method if the token type is fungible.
+        if (keccak256(bytes(tokenType)) == keccak256(bytes("FUNGIBLE_COMMON"))) {
+            result = __redirectForERC20(selector);
+            if (keccak256(result) != keccak256(abi.encode("undefined"))) {
```
acuarica, line 172: In the unlikely case say for example `name` or `symbol` is `"undefined"`, will this work as expected?

@hts-forking/src/slotmap.js
```diff
@@ -189,6 +190,7 @@ function visit(slot, baseSlot, obj, path, map) {
  * @returns {SlotMap}
  */
 function slotMapOf(token) {
+    token['token_type'] = token['type'];
```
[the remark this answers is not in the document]
victor-yanev: https://ethereum.stackexchange.com/questions/84156/is-type-a-reserved-keyword-in-solidity

src/HtsSystemContract.sol
```diff
@@ -160,33 +163,69 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
      */
     function __redirectForToken() internal virtual returns (bytes memory) {
         bytes4 selector = bytes4(msg.data[24:28]);
+        bytes memory result = abi.encode(false);
+        _initTokenData();
+
+        // Redirect to the appropriate ERC20 method if the token type is fungible.
+        if (keccak256(bytes(tokenType)) == keccak256(bytes("FUNGIBLE_COMMON"))) {
```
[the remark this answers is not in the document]
victor-yanev: Good idea, I will try this out

src/HtsSystemContract.sol
victor-yanev: That's right - it wouldn't, perhaps a hash of "undefined" would work better since it's less likely to appear in the actual field

victor-yanev: I will try to emove the "undefined" results and use a modifier for __redirectForHRC719

@hts-forking/src/slotmap.js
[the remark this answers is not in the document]
acuarica: Thanks for the clarification.

contracts/HtsSystemContract.sol
```diff
@@ -212,35 +240,106 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
...
             bool res;
             assembly { res := sload(slot) }
             return abi.encode(res);
-        } else if (selector == this.getTokenInfo.selector) {
+        }
+        revert("redirectForHRC719: not supported");
+    }
+
+    function __redirectForHTS(bytes4 selector) internal virtual returns (bytes memory) {
+        require(msg.sender == HTS_ADDRESS, "redirectForHTS: unauthorized");
```
acuarica, line 330: this was already checked when calling this method, right? If the method is `private` then the double-check is not necessary.

contracts/HtsSystemContract.sol
```diff
@@ -161,32 +164,55 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events {
...
+
+        // Redirect to the appropriate ERC721 method if the token type is non-fungible.
+        if (keccak256(bytes(tokenType)) == keccak256(bytes("NON_FUNGIBLE_UNIQUE"))) {
+            return __redirectForERC721(selector);
+        }
+
+        revert ("redirectForToken: not supported");
+    }
+
+    function __redirectForERC20(bytes4 selector) internal virtual returns (bytes memory) {
```
acuarica, line 187: all these `__redirectFor(ERC20|ERC721|HTS)` should be private.

contracts/HtsSystemContract.sol
```diff
@@ -1,16 +1,19 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
+import {console} from "forge-std/Console.sol";
```
acuarica, line 4: `console`?

contracts/HtsSystemContractJson.sol
```diff
@@ -62,6 +62,13 @@ contract HtsSystemContractJson is HtsSystemContract {
         }
 
         string memory json = mirrorNode().fetchTokenData(address(this));
+        if (vm.keyExistsJson(json, "_status")) {
```
acuarica, line 65: I was thinking here maybe we should set `tokenType = "NOTFOUND"`  or something similar. This would allows us to check whether an address is a token in methods such as `getTokenInfo` or `isToken` (yet to be implemented).

contracts/IERC721.sol
```diff
@@ -0,0 +1,105 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+import {IERC20Events} from "../contracts/IERC20.sol";
```
acuarica, line 4: do we need this?

test/anvil.test.js
```diff
@@ -112,7 +112,8 @@ describe('::anvil', function () {
             name: 'USD Coin',
             symbol: 'USDC',
             decimals: 6n,
-            nrequests: 3,
+            // 4 requests for the `name`, `symbol`, `decimals`, and `tokenType` fields
```
acuarica, line 115:
```suggestion
            // 1 request for each field: `tokenType` (implicit), `name`, `symbol` and `decimals` (4)
            // (both `name` and `symbol` have less than 32 bytes)
```

test/anvil.test.js
```diff
@@ -123,7 +124,9 @@ describe('::anvil', function () {
             name: 'My Crypto Token is the name which the string length is greater than 31',
             symbol: 'Token symbol must be exactly 32!',
             decimals: 0n,
-            nrequests: 7,
+            // 4 requests for the `name`, `symbol`, `decimals`, and `tokenType` fields
+            // (plus 4 from the previous token = 8)
```
acuarica, line 128:
```suggestion
            // 1 request for each field: `tokenType` (implicit), `name`, `symbol` and `decimals` (4)
            // plus (3) for additional `name` words and (1) for additional `symbol` word
```

acuarica (changes requested):

contracts/HtsSystemContract.sol
victor-yanev: That is true, but I think it's better to keep methods independent and not introduce coupling by relying on such assumptions, in the long run it's easier for maintaining and less prone to bugs.

contracts/IERC721.sol
victor-yanev: Nope, it's a leftover - removed it

contracts/HtsSystemContract.sol
victor-yanev: Nvm, I had to remove `redirectForHTS` and inlined it (without reverting if none of the selectors are matched), because we also want to support ERC-20/ERC-721 method calls when the msg.sender is HTS_ADDRESS

src/slotmap.js
```diff
@@ -162,6 +162,7 @@ function visit(slot, baseSlot, obj, path, map) {
         );
     } else if (!label.startsWith('__gap')) {
         const prop = obj[toSnakeCase(label)];
+        if (prop === undefined) return;
```
acuarica, line 165: We should remove this, or even better, we should convert it to an `assert` so we don't mask any error.

contracts/HtsSystemContract.sol
acuarica: "want to support ERC-20/ERC-721 method calls when the msg.sender is HTS_ADDRESS", when does this happen?

victor-yanev: in mint and burn token:
```solidity
        responseCode = 22; // HederaResponseCodes.SUCCESS
        newTotalSupply = int64(uint64(IERC20(token).totalSupply()));
        serialNumbers = new int64[](0);
        require(newTotalSupply >= 0, "mintToken: invalid total supply");
```

```solidity
        responseCode = 22; // HederaResponseCodes.SUCCESS
        newTotalSupply = int64(uint64(IERC20(token).totalSupply()));
        require(newTotalSupply >= 0, "burnToken: invalid total supply");
```

src/slotmap.js
victor-yanev: Test started failing after changing this to use `assert`

victor-yanev: Perhaps, we could only log an error and return safely from `visit`? In this way, we wouldn't break the application and we would still have the error logs for visibility.

acuarica (approved):
Nice work!!

PR #147, opened 2024-12-06
docs: Reestructure readme docs

### LINKED ISSUE #107: Reestructure `README` docs
The current `README` sitting at the project's root folder contains mostly design decisions and implementation details. This is quite useful for contributors to help them understand the project. However, this is not the case for consumers/users of the Hardhat plugin and Foundry library.

Include the appropriate documentation targeting final users in this `README`. This should include basic instructions on how to use both solutions, and any configuration necessary to make it work. The PR[^1] includes some documentation regarding the Foundry library.

One possible solution could be to structure the design docs as following[^2]

- `docs/README.md` - general approach that we've laid out mostly on the `README` in terms of intercepting calls, system contracts, HTS tokens that support ERCs, _etc_.
- `docs/storage.md` - more details and structure to the storage methodology also laid out in the `README`.
- `docs/hardhat-forking.md` - the specific Hardhat approach. The sequence diagram in the `README` belongs here.
- `docs/foundry-forking.md` - the specific Foundry approach. Explain how we use the same base contract to reuse behavior and storage layout, but we provide the actual data from Mirror Node calls within the EVM[^1].

If splitting up this design docs becomes problematic due to how interleaved both solutions are, we could have a single `DESIGN.md` sitting at the root folder containing all of it.

[^1]: https://github.com/hashgraph/hedera-forking/pull/105
[^2]: Thanks @Nana-EC for the suggestion.

acuarica:
**Description**:

This PR makes several improvements to the `README` docs. Now the `README` is focused on end users. It includes instructions and examples on how to use the project with both Foundry and Hardhat.

On the other hand, it adds a new doc `INTERNALS`, that details the background and design decisions made in this project. It describes how the implementation works. The _Storage Layout_ section was moved into a new doc, `INTERNALS`.

Additionally, this PR

- adds NatSpec docs for `IERC20.sol`. This is used to show supported methods in the `README` (see item below).
- adds two new scripts, `abi-table.js` and `readme.js`. The first one displays an ABI table in Markdown format. The second allows to embed the output of commands into `README` (by using a special marker). This two scripts are used together to embed supported HTS method into the `README`.
- includes an HTS example in `examples/foundry-hts` that mimics what an end user would write when using the Foundry library. That's why it uses a different `import` path. To ensure this example is always up to date, a new job was added that runs this test in CI.
- modifies the Prettier configuration to avoid [formatting embedded languages](https://prettier.io/docs/en/options#embedded-language-formatting) in Markdown.

**Related issue(s)**:

Fixes #107.

**Notes for reviewer**:

**Checklist**


mishomihov00 (dismissed):
Review applies to:
`.github/workflows/test.yml`

README.md
```diff
@@ -1,346 +1,523 @@
...
-```console
-forge build
+        string memory name = IERC20(USDC_mainnet).name();
+        string memory symbol = IERC20(USDC_mainnet).symbol();
+        uint8 decimals = IERC20(USDC_mainnet).decimals();
+        assertEq(name, "USD Coin");
+        assertEq(symbol, "USDC");
+        assertEq(decimals, 6);
+
+        console.log("name: %s, symbol: %s, decimals: %d", name, symbol, decimals);
```
quiet-node, line 144: nice! Just wanted to make it clear console.log only works for forking network not on regular local network, correct?

quiet-node: Love to see this tho!

quiet-node (approved):
Great work!!! 🙌🏼

README.md
acuarica: > console.log only works for forking network not on regular local network, correct?

Yes, that's correct.

PR #152, opened 2024-12-11
feat: merge `npm` packages

### LINKED ISSUE #139: [Hardhat] Merge npm packages into a single one
Currently we have two npm packages, `@hts-forking` and `@hardhat-forking-plugin`. The reason behind this is that `@hts-forking` was initially meant to be used from the Relay (related to https://github.com/hashgraph/hedera-forking/issues/122). However, given this is not the case anymore, it might be easier, _e.g._, for publishing or to avoid having symlinked `HtsSystemContract.json`, to have a single npm package. 

If we go this route, we need to restructure the repo, one alternative could the following _(simplified)_ structure

```
.
├── README.md
├── foundry.toml
├── package.json
├── contracts/
│   └── *.sol
├── lib/
│   └── forge-std/
│   └── [more dependencies...]
├── src/
│   └── *.js
└── test/
    └── data/
        └── *.json
    └── *.t.sol
    └── *.test.js
```

this would work, but having `.t.sol` and `.test.js` files in the same folder might seem odd. But on the other hand, both kind of tests use `test/data/` some might make sense to have them in the same folder.

Another issue we need to solve is how to name the package. For example,

```console
$ npm --save-dev install @hashgraph/forking
```

```js
require('@hashgraph/forking/hardhat-plugin');
```

The idea is that you can use the `eth_getCode` and `eth_getStorageAt` even if you're not using Hardhat plugin. That's why I believe it's better is we have it in two different modules. And using only `@hashgraph/forking` (instead of including `hts` somewhere) makes it more general.

acuarica:
**Description**:

This PR merges both `npm` packages, `@hts-forking` and `@hardhat-forking-plugin` into one package sit at the root of the repo. See #139 for the rationale behind this PR.

In particular, this PR

- moves all `.sol` contracts from `src/` into `contracts/`. `foundry.toml` also was updated to reflect this change.
- moves source files from `@hts-forking/src/` into `src/`. Similarly, files from `@hardhat-forking-plugin/src` were moved to `src/plugin/`. Their `package.json` were merged as well. Note the `exports` field, which exports the plugin under `@hashgraph/system-contracts-forking/plugin`.
- moves `@hts-forking/test` into `test/`. Note that this will lead to have `.sol` and `.js` in the same folder. Given that both use the same test data, this is acceptable. Similarly, `@hardhat-forking-plugin/test/` was moved into `test/plugin/`.
- moves test scripts into `test/scripts/` following @victor-yanev suggestion. This includes both `json-rpc-mock` and `curl`.
- fixes the references in `README`.
- adds `blockNumber` to cache response from JSON-RPC in Hardhat plugin tests.

**Related issue(s)**:

Fixes #139.

**Notes for reviewer**:

> [!TIP]
> Even if it looks like a big PR, most of the changes come from moving files around and modifying `require` or `import` paths.

**Checklist**


Nana-EC (approved):
Nice.
Still thinking on alternative to `@hashgraph/system-contracts-forking` but this works for now

PR #157, opened 2024-12-14
test: include e2e to validate HTS emulation against real HTS

### LINKED ISSUE #156: Validate HTS emulation behaves as real HTS for "happy" paths
As we add more functionality to the HTS emulation contracts, we run the risk of differing in behavior from real HTS.

We should include tests that ensure that HTS emulation has the same behavior as real HTS. At least we should focus on "happy" paths, which are the most important ones for users.

acuarica:
**Description**:

This PR includes a validation test to compare the behavior of HTS emulation against real HTS. See `README` for more details on how to run it and an overview of how it works.

It refactor the JS library to export the JSON-RPC forwarder (before was only used by the Hardhat plugin).

Additionally,
- it fixes the script to create tokens by closing the SDK `Client`.
- it includes support to fork from a local node chain `298` in both Foundry library and Hardhat plugin.

**Related issue(s)**:

Fixes #156.

**Notes for reviewer**:

**Checklist**


test/forwarder.test.js
```diff
@@ -49,27 +49,16 @@ class Provider {
...
-                    if (message.listening) {
-                        resolve(message.port);
-                    }
-                })
-                .unref()
-        );
-        provider = new Provider(port);
+        const forkUrl = /**@type{string}*/ (/**@type{unknown}*/ (undefined));
+        const { host } = await jsonRPCForwarder(forkUrl, '', undefined, [
+            '0x70997970c51812dc3a010c7d01b50e0d17dc79c8',
```
natanasow, line 59: Just out of curiosity, what's this address? A random non-existing account or a local node account?

natanasow (approved):

test/forwarder.test.js
acuarica: These are local Hardhat accounts. It's used as an optimization. The forwarder won't forward any rpc call related to these accounts, because it knows they are local hence there should be no remote state to fetch.

Issue #80, opened 2024-10-21
[Hardhat] Implement HTS state emulation for NFT methods


acuarica:
The Hardhat plugin should provide state emulation for NFT methods implemented in https://github.com/hashgraph/hedera-forking/issues/130.

PR #161, opened 2024-12-19
feat: [Foundry] Implement HTS emulation for NFT methods

### LINKED ISSUE #125: [Foundry] Implement HTS state emulation for NFT methods
The Foundry contract should provide state emulation for NFT methods implemented in https://github.com/hashgraph/hedera-forking/issues/130.

victor-yanev:
**Description**:

**Related issue(s)**:

Fixes #125 

**Notes for reviewer**:

**Checklist**


contracts/MirrorNode.sol
```diff
@@ -28,6 +28,37 @@ abstract contract MirrorNode {
 
     function fetchTokenRelationshipOfAccount(string memory account, address token) external virtual returns (string memory json);
 
+    function fetchNonFungibleToken(address token, uint32 serial) external virtual returns (string memory json);
+
+    function getNftMetadata(address token, uint32 serial) external returns (string memory) {
+        try this.fetchNonFungibleToken(token, serial) returns (string memory json) {
```
acuarica, line 34: No need to wrap this in `try/catch`. If there is an error while fetching, I believe it's better to bubble the error up to the user.

Also, if the user has `ffi` disabled, this will mask the error. So better to remove the `try/catch`.

test/ERC721.t.sol
```diff
@@ -36,6 +36,7 @@ contract ERC721TokenTest is Test, TestSetup, IERC721Events, IERC20Events {
         uint192 pad = 0x0;
         bytes32 slot = bytes32(abi.encodePacked(IERC721.ownerOf.selector, pad, uint32(tokenId)));
         vm.store(CFNFTFF, slot, bytes32(uint256(uint160(CFNFTFF_TREASURY))));
+        vm.store(address(bytes20(keccak256(abi.encode(CFNFTFF)))), slot, bytes32(uint(1)));
```
acuarica, line 39: We should remove these `store`s and use mock data for these tests.

contracts/HtsSystemContractJson.sol
```diff
@@ -441,6 +442,42 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+            address approved = mirrorNode().getNftSpender(address(this), serialId);
+            _setValue(slot, bytes32(uint256(uint160(approved))));
+        }
+        return slot;
+    }
+
+    function _isApprovedForAllSlot(address owner, address operator) internal override virtual returns (bytes32) {
+        bytes32 slot = super._isApprovedForAllSlot(owner, operator);
+        if (vm.load(_scratchAddr(), slot) == bytes32(0)) {
+            // TODO: Implement this
```
arianejasuwienas, line 476: @acuarica I'm not entirely sure how to retrieve this information from the Mirrornode, or if it's even possible. :smiling_face_with_tear: . I've been trying to deduce it from the documentation, but I couldn't find anything that matches the 'isApprovedForAll' behavior.

I’ve also been considering intermediary solution:
- Counting all NFTs assigned to a user and then checking how many of those tokens the spender is allowed to spend. If those numbers match, perhaps just returning `true` could work. However, counting them would require listing them first, which means retrieving all pages using the 'next' link...

That said, I’m not sure if this is a good approach, so I wanted to ask for your opinion first. What do you think?

arianejasuwienas:
@acuarica I've moved the allowances implementation to the separate PR. Perhaps I should even create separate issue for that. But clearly somethings seems not to work correctly there. Even though I've set approval through SDK (direct calls to HTS through gRPC) mirronode always returns an empty array when I make call to the allowances/nft endpoint...
https://github.com/hashgraph/hedera-forking/pull/168

arianejasuwienas:
@acuarica I don’t want to remove the draft status yet because some workflows failed. I’m not sure if it’s related to the changes introduced in this PR. Locally, tests are working correctly on both the forked testnet and with the mock files. Moreover, the workflows with ERC20 tests failed, even though those weren’t affected by the changes in this PR. I still think you can review it and let me know if there’s anything else needed at this stage.

I’ve fixed the code to ensure the tests are functional, applied your suggestions from this PR, and merged the latest commits from the main branch.

arianejasuwienas:
Ah, the issue with the workflows was on my side. Support for the new fields needs to be added to the JS server (src/index.js). I'm working on it now.

contracts/Base64.sol
```diff
@@ -0,0 +1,71 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity >=0.6.0;
+
+bytes constant TABLE_DECODE = hex"0000000000000000000000000000000000000000000000000000000000000000"
+hex"00000000000000000000003e0000003f3435363738393a3b3c3d000000000000"
+hex"00000102030405060708090a0b0c0d0e0f101112131415161718190000000000"
+hex"001a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132330000000000";
+
+function decode(string memory _data) pure returns (bytes memory) {
```
arianejasuwienas, line 10: Maybe it would be better to use Solady instead? @acuarica

test/scripts/blocks_13890397.json
```diff
@@ -3,7 +3,7 @@
     "hapi_version": "0.52.0",
     "hash": "0x7e8a7062f9b66e76c8f267f19a4db60742f63041ac9ebb9b835c8150191e37375d726554dbeae40eccfd0792df7d267b",
     "name": "2024-08-30T13_25_28.756521003Z.rcd.gz",
-    "number": 8535327,
+    "number": 13890397,
```
acuarica, line 6: We should either leave the `number` unchanged, or replace the block data altogether. Otherwise, it might be confusing where this file came from.

test/scripts/curl
```diff
@@ -46,6 +46,14 @@ for (const [re, fn] of /** @type {const} */([
 
         return require(`../data/${token.symbol}/getBalanceOfToken_${accountId}.json`);
     }],
+    [/^tokens\/([\w.]+)\/nfts\/(\d+)$/, (tokenId, nftNumber) => {
+        assert(typeof tokenId === 'string');
+        const token = tokens[tokenId];
+        if (token === undefined)
+            return { _status: { messages: [{ message: 'Not found' }] } };
```
acuarica, line 53: We should return `undefined` here, so the main loop then return this not found response.

acuarica (commented):
Looking good. Please update the PR description to mention the most relevant changes. In particular, why it was needed to change the block number.

Moreover, these changes include also the JS/Hardhat implementation, right? There is another issue for that https://github.com/hashgraph/hedera-forking/issues/80. Maybe we should split this PR into two to cover both cases. Otherwise it's difficult to review.

mishomihov00 (approved):
Review applies to `.github/workflows/test.yml`

test/scripts/blocks_13890397.json
arianejasuwienas: @acuarica I need to keep the current solution and change the block number. This is because, at the previous block number, my NFT (with a tokenUri in the metadata exceeding 31 bytes) had not yet been deployed. The same applies to the approvals that were prepared for testing purposes.

arianejasuwienas:
@acuarica As requested, this PR has been replaced with two separate PRs: one for Foundry and one for the JS Library. I’ll close this PR but won’t delete the branch yet, as I’m not entirely sure the currently proposed solution is better.

PR #162, opened 2024-12-19
feat: hedera token service embedding (#118)

### LINKED ISSUE #118: [Foundry] Investigate HTS "embedding" in Foundry
HTS emulation as currently implemented posses two main issues

1. **Fully re-implement HTS behavior.** HTS contains many validations and customizations, which can be difficult to get right in Solidity.
2. **Maintenance.** Even if we manage to fully replicate HTS behavior (and that’s a big IF) in EVM bytecode, we have now to essentially maintain two implementation of HTS: The one in services and the one implemented here. This can be error prone and cumbersome to maintain.

An alternative to HTS _"emulation"_ could be HTS _"embedding”_. The idea is to _somehow_, embed `hedera-services` behavior into Foundry or Hardhat. Given this approach would be tailored to each dev environment, we will probably end up writing one “adapter” for each dev env we want to provide forking support for.

### Possible solution for Foundry

Foundry allows users to call an arbitrary executable through their [`ffi`](https://book.getfoundry.sh/cheatcodes/ffi) cheatcode. We could, _in principle_, build a custom `hedera-services` executable that only contains HTS System Contract related behavior. This can be an application that takes the `calldata` in the command line and returns the appropriate value.

There are some issues that we need to solve in order to implement this approach

```[tasklist]
### Tasks
- [ ] **1.** How to fetch remote state into the HTS process.
- [ ] **2.** Ensure HTS state is isolated from the rest of the (Foundry's) EVM state.
- [ ] **3.** How to maintain HTS state between calls. Foundry does not have a hook to kill a process when the tests finish, meaning we cannot have a running service. Which implies that each call to HTS would invoke a new instance of the executable. Keep in mind that tests [don't share state](https://book.getfoundry.sh/forge/writing-tests#before-test-setups), meaning each call to HTS could use different states.
- [ ] **4.** Create a native executable embedding the JVM and HTS from `hedera-services`. This should be as lightweight as possible, given it would be invoked on each HTS call.
- [ ] **5.** Ensure that HTS behavior related classes **do not** depend on Besu EVM. Besu shouldn't be needed because HTS does not run EVM bytecode, only Java code.
```

image

### PoC

```[tasklist]
- [ ] Investigate how the Mirror Node simulates `eth_call` requests. This might give us an idea how to embed and extract `hedera-services` for our purposes.
```

### Related Issues & Discussions

- https://github.com/hashgraph/hedera-local-node/discussions/252

### Related Work

Just for reference, Arbitrum also uses “precompiles” or System Contracts to provide additional functionality https://docs.arbitrum.io/build-decentralized-apps/precompiles/reference. And it also suffers from the same problem as us regarding forking https://docs.arbitrum.io/for-devs/troubleshooting-building#why-does-my-transaction-revert-with-invalidfeopcode-when-using-foundry

> - **Why does my transaction revert with InvalidFEOpcode when using Foundry?**
>    
>   Foundry and other similar development tools that enable chain forking, do not support Arbitrum precompiles. If your transaction is calling a precompile, it is likely that it will revert with `InvalidFEOpcode`.
>    
>   To rule out that possibility, it is recommended to send the transaction with a different tool.

#### Foundry

Arbitrum users have similar issues regarding forking, for example https://github.com/foundry-rs/foundry/issues/5085.

It seems `zskSync` opts for custom Foundry support https://github.com/foundry-rs/foundry/issues/7624. We may want to implement a solution similar to https://github.com/matter-labs/foundry-zksync.

There is an ongoing effort in Foundry to support precompiles (in the context of EVM-compatible networks with additional functionality)

- https://github.com/foundry-rs/foundry/issues/748
- https://github.com/foundry-rs/foundry/issues/8266
- https://github.com/foundry-rs/foundry/issues/1505
- https://github.com/foundry-rs/foundry/issues/7498
- https://github.com/foundry-rs/foundry/issues/7262

However, it does not look like allows external code to injected yet https://github.com/foundry-rs/foundry/pull/7589. Only available when using Anvil as a library.

#### Hardhat

Similar discussions to support EVM-compatible networks with additional semantics also happen in Hardhat https://github.com/NomicFoundation/hardhat/issues/1902.

arianejasuwienas:
**Description**:
Sample usage of the embedded HTS service in the foundry tests using FFI. For now only PoC using the already running Hedera Local node is supported.

**Related issue(s)**:

Fixes #118.

**Notes for reviewer**:

**Checklist**


test/ffi/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "ffi",
+    "version": "1.0.0",
+    "main": "createToken.js",
```
acuarica, line 4: is this file missing?

acuarica:
As discussed, the idea of having this PoC as a PR is to not lose progress, in case we want to resume it in the future.

PR #173, opened 2025-01-07
feat: [Hardhat] implement nft support for the json rpc api library and js mock …

arianejasuwienas:
**Description**:
Implementing support for NFTs in the JS library for the JSON-RPC mock, as well as in the JSON-RPC server that utilizes this library.

#### The reason I've used persistent storage here:
We cannot determine which NFT numbers have been minted unless someone specifically queries them. This can be addressed in one of two ways:

1. Dynamic loading on first access: Load each NFT's details dynamically the first time its slot is queried. Once accessed, store information about the existence of this NFT. This is my current approach. However, this may result in a memory leak/optimisation when a large number of tests or NFT requests are made, as memory is never freed (this is where the name: persistentStorage comes from).

2. Preloading All NFTs: Load the details of all NFTs each time the token storage is accessed. While this avoids memory leaks, it could lead to excessive memory usage and generate a large number of unnecessary API requests for NFTs that we are not interested about :/ .

#### Why it might be an issue:
Let's consider scenario where:
1. We have a long tokenUri that spans more than one storage slot. 
2. Someone attempts to access the value in the second slot without first querying the initial slot.

We won’t be able to provide the correct result. This is because persistentStorage is only populated when the request for the first slot of the string is made (only then we can deduce the serial NFT number from the slot address). I don't think it's not a critical issue though...

**Related issue(s)**:

#80

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ const { strict: assert } = require('assert');
 const debug = require('util').debuglog('hts-forking');
 
 const { ZERO_HEX_32_BYTE, toIntHex256 } = require('./utils');
-const { slotMapOf, packValues } = require('./slotmap');
+const { slotMapOf, packValues, persistentSlotMapOf } = require('./slotmap');
 const { deployedBytecode } = require('../out/HtsSystemContract.sol/HtsSystemContract.json');
 
 const HTSAddress = '0x0000000000000000000000000000000000000167';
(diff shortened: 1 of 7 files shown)


src/slotmap.js
```diff
@@ -274,4 +274,60 @@ function slotMapOf(token) {
...
+ *   path:string;
+ *   type:string;
+ * }>|undefined} load - Loads the value associated with a slot.
+ * @property {function(bigint, string): void} store - Stores a value in the slot map.
+ */
+
+/**
+ * @type {Array<PersistentStorage>}
+ */
+const persistentStorage = [];
```
acuarica, line 305: this global state can bring problems in the future. Given this is an optimization, we should remove it for now.

src/index.js
```diff
@@ -138,11 +138,118 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
...
+            metadata: null,
+        };
+        if (typeof metadata !== 'string')
+            return ret(
+                ZERO_HEX_32_BYTE,
+                `Failed to get the metadata of the NFT ${tokenId}#${serialId}`
+            );
+        persistentSlotMapOf(tokenId).store(nrequestedSlot, atob(metadata));
+    }
+    let unresolvedValues = persistentSlotMapOf(tokenId).load(nrequestedSlot);
```
acuarica, line 244: moreover, not sure this works as intended, since it's going to always write when asked for token uri.

test/data/CFNFTFF/getAllowanceForToken_0.0.4436613_0.0.4454450.json
```diff
@@ -0,0 +1,19 @@
+{
```
acuarica, line 1: I'm guessing all mock data files are the same as the Foundry one's, so this should be already included in the Foundry PR, right?

acuarica (changes requested):
looking good, left some questions/comments

test/data/CFNFTFF/getAllowanceForToken_0.0.4436613_0.0.4454450.json
arianejasuwienas: Yes, this will be already included in the previous PR

src/slotmap.js
arianejasuwienas: @acuarica 
The issue is that we cannot determine which NFT numbers have been minted unless someone specifically queries them. This can be addressed in one of two ways:

1. Dynamic loading on first access: Load each NFT's details dynamically the first time its slot is queried. Once accessed, store information about the existence of this NFT. This is my current approach. However, this may result in a memory leak/optimisation when a large number of tests or NFT requests are made, as memory is never freed (this is where the name: persistentStorage comes from).

2. Preloading All NFTs: Load the details of all NFTs each time the token storage is accessed. While this avoids memory leaks, it could lead to excessive memory usage and generate a large number of unnecessary API requests for NFTs that we are not interested about :/ .

src/index.js
arianejasuwienas: When querying the first slot holding the tokenUri, it will store the values for all slots containing this single string.

src/slotmap.js
acuarica: > unless someone specifically queries them.

That's fine. Why is this a problem?

acuarica:
To improve the discussion regarding `persistentSlotMapOf`, please describe how it works in the PR.

src/slotmap.js
arianejasuwienas: @acuarica Problematic sceario:
1. We have a long tokenUri that spans more than one storage slot. 
2. Someone attempts to access the value in the second slot without first querying the initial slot.

We won’t be able to provide the correct result. This is because persistentStorage is only populated when the request for the first slot of the string is made (only then we can deduce the serial NFT number from the slot address). I don't think it's not a critical issue though...

src/slotmap.js
```diff
@@ -274,4 +274,60 @@ function slotMapOf(token) {
...
+
+/**
+ * Slot map of token, but persistent, will not be removed between multiple separate requests.
+ *
+ * @param {string} target
+ * @returns {PersistentSlotMap} An object with `load` and `store` methods.
+ */
+function persistentSlotMapOf(target) {
+    const found = persistentStorage.filter(occupiedSlot => occupiedSlot.target === target);
```
acuarica, line 314: what about `blockNumber`? The key should be `target` and `blockNumber` because the user can change the `blockNumber`.

mishomihov00 (approved):
Review applies to `.github/workflows/test.yml`

arianejasuwienas:
@acuarica Thank you for the review! I’ve applied your suggestions - please let me know if everything looks good now.

src/slotmap.js
```diff
@@ -274,4 +274,54 @@ function slotMapOf(token) {
     return map;
 }
 
-module.exports = { slotMapOf, packValues };
+/**
+ * Represents the value in the persistent storage.
+ * Each token has its own SlotMap. Any value can be assigned to this storage
```
acuarica, line 279:
```suggestion
 * Each token and `blockNumber` has its own SlotMap. Any value can be assigned to this storage
```

src/slotmap.js
```diff
@@ -274,4 +274,54 @@ function slotMapOf(token) {
...
+     * @param {string} tokenId
+     * @param {number} blockNumber
+     */
+    _init(tokenId, blockNumber) {
+        const key = `${tokenId}:${blockNumber}`;
+        const initialized = this._map.get(key) || new SlotMap();
+        if (!this._map.has(key)) {
+            this._map.set(key, initialized);
+        }
+        return initialized;
```
acuarica, line 298: to avoid the double guard

```suggestion
        let slotMap = this._map.get(key);
        if (slotMap === undefined) {
            slotMap = new SlotMap();
            this._map.set(key, slotMap);
        }        
        return slotMap;
```

src/slotmap.js
```diff
@@ -274,4 +274,54 @@ function slotMapOf(token) {
...
+            '',
+            this._init(tokenId, blockNumber)
+        );
+    }
+
+    /**
+     * @param {string} tokenId
+     * @param {number} blockNumber
+     * @param {bigint} slot
+     */
```
acuarica, line 321: nit: let's add return type for clarity.

src/slotmap.js
```diff
@@ -274,4 +274,54 @@ function slotMapOf(token) {
...
+    constructor() {
+        /** @type {Map<string, SlotMap>} */
+        this._map = new Map();
+    }
+
+    /**
+     * @param {string} tokenId
+     * @param {number} blockNumber
+     */
+    _init(tokenId, blockNumber) {
```
acuarica, line 292:
```suggestion
    _getSlotMap(tokenId, blockNumber) {
```

acuarica (commented):
looking good, final touches

src/slotmap.js
```diff
@@ -274,4 +274,54 @@ function slotMapOf(token) {
     return map;
 }
 
-module.exports = { slotMapOf, packValues };
+/**
+ * Represents the value in the persistent storage.
```
acuarica, line 278: There is some information in the PR description that would be really useful here. In particular **1.** of section "The reason I've used persistent storage here" and section "Why it might be an issue".

We should include that info as a JS doc here.

arianejasuwienas:
@acuarica I've applied all your suggestions. Let me know if there's anything else worth improving :) .

acuarica (approved):
Great stuff!

Issue #151, opened 2024-12-11
Implement `IHederaTokenService` facade methods


acuarica:
The [`IHederaTokenService`](https://github.com/hashgraph/hedera-smart-contracts/blob/b29a7dc6969f2ca5809d1da265d0ff68f4a62bf2/contracts/system-contracts/hedera-token-service/IHederaTokenService.sol) interface contains many facade methods that are already implemented in `HtsSystemContract`, _e.g._, transfer, approvals and association methods.

Extend our `IHederaTokenService` interface and implement these methods similar to `getTokenInfo`.

acuarica:
@arianejasuwienas this should be closed. Please link it or mention the PR where this was implemented so we can close this.

arianejasuwienas:
@acuarica I've linked the PR: https://github.com/hashgraph/hedera-forking/pull/188

PR #188, opened 2025-01-14
feat: methods for transfer association and approval from the IHTS

arianejasuwienas:
**Description**:

Adding the implementation of the methods  from the IHederaTokenService interface. I will allow users to perform token-based operations by interacting with the HTS contract at the 0x167 address. The token-based operations introduced in this PR were already implemented on the token side through interfaces compatible with IERC721, IERC20, and IHRC719, but they have now been extended to support HTS calls as well.

For example, in addition to transferring a fungible token using the IERC20 interface

```solidity
IERC20(token).transferFrom(sender, recipient, amount);  
```

you can now perform the same operation via the HTS interface

```solidity
IHederaTokenService(address(0x167)).transferToken(token, sender, recipient, amount);
```

**Related issue(s)**:

#151

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -238,33 +238,27 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
     function getTokenCustomFees(
         address token
     ) htsCall external returns (int64, FixedFee[] memory, FractionalFee[] memory, RoyaltyFee[] memory) {
-        require(token != address(0), "getTokenCustomFees: invalid token");
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         return (responseCode, tokenInfo.fixedFees, tokenInfo.fractionalFees, tokenInfo.royaltyFees);
     }
 
     function getTokenDefaultFreezeStatus(address token) htsCall external returns (int64, bool) {
-        require(token != address(0), "getTokenDefaultFreezeStatus: invalid address");
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         return (responseCode, tokenInfo.token.freezeDefault);
     }
 
     function getTokenDefaultKycStatus(address token) htsCall external returns (int64, bool) {
-        require(token != address(0), "getTokenDefaultKycStatus: invalid address");
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         return (responseCode, tokenInfo.defaultKycStatus);
     }
 
     function getTokenExpiryInfo(address token) htsCall external returns (int64, Expiry memory expiry) {
-        require(token != address(0), "getTokenExpiryInfo: invalid token");
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         return (responseCode, tokenInfo.token.expiry);
     }
 
     function getFungibleTokenInfo(address token) htsCall external returns (int64, FungibleTokenInfo memory) {
-        require(token != address(0), "getFungibleTokenInfo: invalid token");
-
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         require(responseCode == HederaResponseCodes.SUCCESS, "getFungibleTokenInfo: failed to get token data");
         FungibleTokenInfo memory fungibleTokenInfo;
         fungibleTokenInfo.tokenInfo = tokenInfo;
@@ -273,15 +267,14 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         return (responseCode, fungibleTokenInfo);
     }
 
-    function getTokenInfo(address token) htsCall external returns (int64, TokenInfo memory) {
+    function getTokenInfo(address token) htsCall public returns (int64, TokenInfo memory) {
         require(token != address(0), "getTokenInfo: invalid token");
 
         return IHederaTokenService(token).getTokenInfo(token);
     }
 
     function getTokenKey(address token, uint keyType) htsCall external returns (int64, KeyValue memory) {
-        require(token != address(0), "getTokenKey: invalid token");
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         require(responseCode == HederaResponseCodes.SUCCESS, "getTokenKey: failed to get token data");
         for (uint256 i = 0; i < tokenInfo.token.tokenKeys.length; i++) {
             if (tokenInfo.token.tokenKeys[i].keyType == keyType) {
@@ -295,9 +288,7 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
     function getNonFungibleTokenInfo(address token, int64 serialNumber)
         htsCall external
         returns (int64, NonFungibleTokenInfo memory) {
-        require(token != address(0), "getNonFungibleTokenInfo: invalid token");
-
-        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         require(responseCode == HederaResponseCodes.SUCCESS, "getNonFungibleTokenInfo: failed to get token data");
         NonFungibleTokenInfo memory nonFungibleTokenInfo;
         nonFungibleTokenInfo.tokenInfo = tokenInfo;
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -189,6 +189,10 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         return transferToken(token, sender, recipient, int64(int256(amount)));
     }
 
+    function allowance(address token, address owner, address spender) htsCall external view returns (int64, uint256) {
+        return (HederaResponseCodes.SUCCESS, IERC20(token).allowance(owner, spender));
+    }
+
     function approveNFT(
         address token,
         address approved,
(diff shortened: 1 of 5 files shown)


contracts/IHederaTokenService.sol
```diff
@@ -305,7 +305,7 @@ interface IHederaTokenService {
     /// @param tokenTransfers the list of token transfers to do
     /// @custom:version 0.3.0 the signature of the previous version was cryptoTransfer(TokenTransferList[] memory tokenTransfers)
     // function cryptoTransfer(TransferList memory transferList, TokenTransferList[] memory tokenTransfers)
-    //     external
+    //     payable external
```
acuarica, line 308: remove for now, but why did you have to add in the first place?

contracts/HtsSystemContract.sol
```diff
@@ -5,6 +5,7 @@ import {IERC20Events, IERC20} from "./IERC20.sol";
 import {IERC721, IERC721Events} from "./IERC721.sol";
 import {IHRC719} from "./IHRC719.sol";
 import {IHederaTokenService} from "./IHederaTokenService.sol";
+import {IERC165} from "../lib/forge-std/src/interfaces/IERC165.sol";
```
acuarica, line 8: is it used? if not, we should remove it

contracts/HtsSystemContract.sol
```diff
@@ -48,6 +49,265 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         (responseCode, tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
     }
 
+    function getNonFungibleTokenInfo(address token, int64 serialNumber)
+        htsCall external
+        returns (int64, NonFungibleTokenInfo memory) {
+        require(token != address(0), "getNonFungibleTokenInfo: invalid token");
+
+        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        require(responseCode == 22, "getNonFungibleTokenInfo: failed to get token data");
```
acuarica, line 58: let's extract this to a library like https://github.com/hashgraph/hedera-smart-contracts/blob/ffe75e67cbf4011b561be571b5747251cbb60dad/contracts/system-contracts/HederaResponseCodes.sol#L31 with only `HederaResponseCodes.SUCCESS` and replace any other usage

contracts/HtsSystemContract.sol
```diff
@@ -48,6 +49,265 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         (responseCode, tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
     }
 
+    function getNonFungibleTokenInfo(address token, int64 serialNumber)
```
acuarica, line 52: nit: let's sort methods as they are declared in `IHederaTokenService`. This will improve readability, as the same time it will group related methods together, _e.g._, `associateTokens` and `associateToken`.

contracts/HtsSystemContract.sol
```diff
@@ -48,6 +49,265 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+
+        return (responseCode, fungibleTokenInfo);
+    }
+
+    function associateTokens(address account, address[] memory tokens) htsCall public returns (int64 responseCode) {
+        require(tokens.length > 0, "associateTokens: missing tokens");
+        require(account == msg.sender, "associateTokens: Must be signed by the provided Account's key or called from the accounts contract key");
+        for (uint256 i = 0; i < tokens.length; i++) {
+            require(tokens[i] != address(0), "associateTokens: invalid token");
+            int64 associationResponseCode = IHederaTokenService(tokens[i]).associateToken(account, tokens[i]);
```
acuarica, line 89: can't we use `IHRC719(tokens[i]).associateToken()` instead? this will avoid having an extra entry in `redirectForToken`.

contracts/HtsSystemContract.sol
```diff
@@ -48,6 +49,265 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+            from = recipient;
+            to = sender;
+            amount *= -1;
+        }
+        require(
+            from == msg.sender ||
+            IERC20(token).allowance(from, msg.sender) >= uint256(uint64(amount)),
+            "transferNFT: unauthorized"
+        );
+        HtsSystemContract(token)._transferAsHTS(from, to, uint256(uint64(amount)));
```
acuarica, line 170: can't we use `IERC20(token).transferFrom` directly? the same validations should be there, so we don't duplicate ourselves.

test/HTS.t.sol
```diff
@@ -303,4 +305,479 @@ contract HTSTest is Test, TestSetup {
...
+        // https://hashscan.io/testnet/account/0.0.1421
+        address owner = 0x4D1c823b5f15bE83FDf5adAF137c2a9e0E78fE15;
+        address to = makeAddr("bob");
+        uint256 amount = 4_000000;
+
+        uint256 balanceOfOwner = IERC20(USDC).balanceOf(owner);
+        assertGt(balanceOfOwner, 0);
+        assertEq(IERC20(USDC).balanceOf(to), 0);
+
+        vm.prank(owner); // https://book.getfoundry.sh/cheatcodes/prank
```
acuarica, line 623:
```suggestion
        vm.prank(owner);
```

acuarica (changes requested):
nice work, it's looking good

contracts/HtsSystemContract.sol
arianejasuwienas: @acuarica This operation would be triggered by the HTS address rather than the original caller (we would have msg.sender == 0x167). We need to forward the original sender’s address.

contracts/HtsSystemContract.sol
arianejasuwienas: @acuarica For similar reasons I've described here we can't entirely remove additional redirectForToken, but I managed to make it working without new methods with AsHTS suffix, by editing already implemented _approve / _transfer private methods (I've added the sender as an input argument).

Let me know if that will work for us.

contracts/IHederaTokenService.sol
arianejasuwienas: @acuarica To ensure HBAR transfers work correctly. This will work only in scenarios where msg.sender transfers the entire input balance (msg.value) directly.

But yes, it is not needed in this PR.

contracts/HtsSystemContract.sol
```diff
@@ -78,18 +73,249 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+
+    function transferFrom(
+        address token,
+        address sender,
+        address recipient,
+        uint256 amount
+    ) htsCall external returns (int64) {
+        return transferToken(token, sender, recipient, int64(int256(amount)));
+    }
+
```
acuarica, line 191: can we include `allowance` as well? this shouldn't create an entry in redirectForToken given no `msg.sender` is used, right?

contracts/HtsSystemContract.sol
```diff
@@ -78,18 +73,249 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+    function isApprovedForAll(
+        address token,
+        address owner,
+        address operator
+    ) htsCall external view returns (int64, bool) {
+        require(token != address(0), "isApprovedForAll: invalid token");
+        return (HederaResponseCodes.SUCCESS, IERC721(token).isApprovedForAll(owner, operator));
+    }
+
+    function getTokenCustomFees(
```
acuarica, line 234: let's use `getTokenInfo` directly to avoid an entry in `redirectForToken`

contracts/HtsSystemContract.sol
```diff
@@ -78,18 +73,249 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        require(token != address(0), "getTokenCustomFees: invalid token");
+        return IHederaTokenService(token).getTokenCustomFees(token);
+    }
+
+    function getTokenDefaultFreezeStatus(address token) htsCall external returns (int64, bool) {
+        require(token != address(0), "getTokenDefaultFreezeStatus: invalid address");
+        return IHederaTokenService(token).getTokenDefaultFreezeStatus(token);
+    }
+
+    function getTokenDefaultKycStatus(address token) htsCall external returns (int64, bool) {
```
acuarica, line 246: let's use `getTokenInfo` directly to avoid an entry in `redirectForToken`

contracts/HtsSystemContract.sol
```diff
@@ -78,18 +73,249 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+    }
+
+    function getTokenCustomFees(
+        address token
+    ) htsCall external returns (int64, FixedFee[] memory, FractionalFee[] memory, RoyaltyFee[] memory) {
+        require(token != address(0), "getTokenCustomFees: invalid token");
+        return IHederaTokenService(token).getTokenCustomFees(token);
+    }
+
+    function getTokenDefaultFreezeStatus(address token) htsCall external returns (int64, bool) {
```
acuarica, line 241: let's use `getTokenInfo` directly to avoid an entry in `redirectForToken`

contracts/HtsSystemContract.sol
```diff
@@ -78,18 +73,249 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        require(token != address(0), "getTokenDefaultFreezeStatus: invalid address");
+        return IHederaTokenService(token).getTokenDefaultFreezeStatus(token);
+    }
+
+    function getTokenDefaultKycStatus(address token) htsCall external returns (int64, bool) {
+        require(token != address(0), "getTokenDefaultKycStatus: invalid address");
+        return IHederaTokenService(token).getTokenDefaultKycStatus(token);
+    }
+
+    function getTokenExpiryInfo(address token) htsCall external returns (int64 responseCode, Expiry memory tokenInfo) {
```
acuarica, line 251: let's use `getTokenInfo` directly to avoid an entry in `redirectForToken`

acuarica (commented):
looking good, almost there

contracts/HtsSystemContract.sol
arianejasuwienas: Yes, I just missed this method, there was no reason not to include it at this stage, so I added it....

contracts/HtsSystemContract.sol
arianejasuwienas: Is the current solution ok?

Now we have less scenarios to support in the second call to the `redirectForToken`.

But there will be always 1 additional enterance to the `redirectForToken`. This won't change. We are in the context of the `HTS`, and we want to access the token memory, so we have to forward the request from the HTS to the token. Even `getTokenInfo` method looks like this (for the same reason):

```
    function getTokenInfo(address token) htsCall public returns (int64, TokenInfo memory) {
        require(token != address(0), "getTokenInfo: invalid token");

        return IHederaTokenService(token).getTokenInfo(token);
    }
    ```

File: contracts/HtsSystemContract.sol [hunk 13]
```diff
@@ -78,18 +73,256 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        KeyValue memory emptyKey;
+        return (HederaResponseCodes.SUCCESS, emptyKey);
+    }
+
+    function getNonFungibleTokenInfo(address token, int64 serialNumber)
+        htsCall external
+        returns (int64, NonFungibleTokenInfo memory) {
+        require(token != address(0), "getNonFungibleTokenInfo: invalid token");
+
+        (int64 responseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
```
acuarica, line 300:
```suggestion
        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
```

this way we can reuse the validation from `getTokenInfo`

acuarica (approved):
Nice stuff!

Issue #191, opened 2025-01-14
Include ERC721 methods in `examples`


acuarica:
Include examples on how to use ERC721 methods to showcase NFT features.

PR #209, opened 2025-01-21
feat: example usage in nft testing of our foundry library and hardhat plugin (#191)

arianejasuwienas:
**Description**:

1. Basic usage of our Foundry library and Hardhat plugin has been added for NFTs. Users can now explore how our solution can be utilized in examples.
2. Adding support for royalties in both hardhat and foundry (it did not work).

**Related issue(s)**:

#191

**Notes for reviewer**:

**Checklist**


src/utils.js
```diff
@@ -51,5 +51,10 @@ module.exports = {
      * @param {string | number | bigint} value It can be anything accepted by the [`BigInt` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/BigInt).
      * @returns {string}
      */
-    toIntHex256: value => BigInt(value).toString(16).padStart(64, '0'),
+    toIntHex256: value => {
+        if (!['string', 'number', 'bigint'].includes(typeof value)) {
```
arianejasuwienas, line 55: We had an error here... Sometimes, when the balance was big enough this method was already invoked with an object (BigInt) instead of the expected types..

src/forwarder/mirror-node-client.js
```diff
@@ -65,12 +65,11 @@ class MirrorNodeClient {
...
      * @param {string} tokenId the token ID to fetch.
      * @param {number} serialId the serial ID of the NFT to fetch.
-     * @param {number} blockNumber
+     * @param {number} _blockNumber
      * @returns {Promise<Record<string, unknown> | null>} a `Promise` resolving to the token information or null if not found.
      */
-    async getNftByTokenIdAndSerial(tokenId, serialId, blockNumber) {
-        const timestamp = await this.getBlockQueryParam(blockNumber);
-        return this._get(`tokens/${tokenId}/nft/${serialId}?${timestamp}`);
+    async getNftByTokenIdAndSerial(tokenId, serialId, _blockNumber) {
```
arianejasuwienas, line 71: 1. Previous url was not correct.
2. It does not support the block numbers ;/ . That is why I'm not checking if the token belongs to any specific account in my tests - it can change the ownership in time.

arianejasuwienas:
A lot of tests started to fail because of the:

```
     Error: the string "Warning: This is a nightly build of Foundry. It is recommended to use the latest stable version. Visit https://book.getfoundry.sh/announcements for more information. \nTo mute this warning set `FOUNDRY_DISABLE_NIGHTLY_WARNING` in your environment. \n\n" was thrown, throw an Error :)
```
Should we set this env variable inside the runner? Or update the foundry version used in the runner?

acuarica:
> Or update the foundry version used in the runner?

That would be better. However, we need to wait until the actions we use support the `stable` version.

src/slotmap.js
```diff
@@ -267,7 +267,17 @@ function slotMapOf(token) {
         maximum_amount: fee['maximum'],
         fee_collector: fee['collector_account_id'],
     }));
-    token['royalty_fees'] = customFees['royalty_fees'] ?? [];
+    token['royalty_fees'] = (customFees['royalty_fees'] ?? []).map(fee => ({
```
arianejasuwienas, line 270: Royalty fees had to be correctly mapped from the mirrornode response api structure into the Solidity structure. It used to fail otherwise.

contracts/HtsSystemContractJson.sol
```diff
@@ -201,6 +201,16 @@ contract HtsSystemContractJson is HtsSystemContract {
         return tokenInfo;
     }
 
+    // In order to properly decode the bytes returned by the parseJson into the Solidity Structure, the full,
+    // correct structure has to be provided in the input json, with all of the corresponding fields.
+    function _sanitizeFeesStructure(string memory json) private pure returns (string memory) {
```
arianejasuwienas, line 206: fallback_fee is nullable in the json response from the mirror node, but abi decode requires it to be set in order to create proper structure. That is why I set the default value here when it is empty.

examples/hardhat-hts/test/nft-ownerof.test.js
```diff
@@ -0,0 +1,31 @@
...
+const { expect } = require('chai');
+// prettier-ignore
+const { ethers: { getContractAt } } = require('hardhat');
+
+describe('NFT example -- ownerOf', function () {
+    it('should get `ownerOf` account holder', async function () {
+        // https://hashscan.io/mainnet/token/0.0.4970613
+        const nft = await getContractAt('IERC721', '0x00000000000000000000000000000000004bd875');
+        expect(await nft['ownerOf'](1)).to.not.be.equal(
+            '0x0000000000000000000000000000000000000000'
```
acuarica, line 28: can we compare against the current owner?

examples/hardhat-hts/test/nft.test.js
```diff
@@ -0,0 +1,126 @@
...
+        const owner = await ethers.getSigner(ownerAddress);
+        expect(await nft['ownerOf'](serialId)).to.be.equal(owner.address);
+
+        const CallToken = await ethers.getContractFactory('CallNFT');
+        const callToken = await CallToken.deploy();
+        const callTokenAddress = await callToken.getAddress();
+
+        await connectAs(nft, owner)['approve'](callTokenAddress, serialId);
+        await connectAs(callToken, spender)['invokeTransferFrom'](nftAddress, receiver, serialId);
+        expect(await nft['getApproved'](serialId)).to.not.be.equal(spender.address);
```
acuarica, line 123: can we compare for the positive?

examples/hardhat-hts/package-lock.json
```diff
@@ -870,8 +870,8 @@
             }
         },
         "node_modules/@hashgraph/system-contracts-forking": {
-            "version": "0.0.1",
-            "resolved": "git+ssh://git@github.com/hashgraph/hedera-forking.git#daa80f304e396b35ab84135176ac6e48b8cab35f",
+            "version": "0.1.1",
+            "resolved": "git+ssh://git@github.com/hashgraph/hedera-forking.git#8533630770865fc5e3ae0341aee2b71b7d229350",
```
acuarica, line 874: let's install from npm (instead of git) so it's closer to a what an actual user would use it.

contracts/HtsSystemContractJson.sol
acuarica: sure, but also there is a `vm.parseJson(json, ".custom_fees.royalty_fees")` that guards against `null`, right?

arianejasuwienas: @acuarica Only when royalty_fees is null. For the barkaneer I had fallback_fee = null (substructure in the royalty fees object).

examples/hardhat-hts/test/nft.test.js
arianejasuwienas: Yes, we can. But we get nft details for the latest block, so all it will take for our tests then is someone who owns this serial id = 1 to transfer it to someone else. (I'll apply your suggestion anyway :) ).

examples/hardhat-hts/test/nft-ownerof.test.js
arianejasuwienas: Yes, we can. But we get nft details for the latest block, so all it will take for our tests then is someone who owns this serial id = 1 to transfer it to someone else. (I'll apply your suggestion anyway :) ).

examples/hardhat-hts/package-lock.json
arianejasuwienas: Ok, I am getting it directly from npm, @acuarica . Unfortunately current version (0.1.1) won't work for this token anyway. Because it has royalty fees (and because of the incorrrect nft details path).

Perhaps we should split it in 2 PRs.

First: https://github.com/hashgraph/hedera-forking/pull/218

And then, when the library is fixed we can update this package lock. 

OR

We can merge this PR and create a new one soon after, with updated dependency (it's up to you). I like this idea even more, because then we would have the fix in the solidity smart contract too (for nullable substructure in royalty fees).

acuarica (approved):

contracts/HtsSystemContractJson.sol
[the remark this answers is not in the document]
arianejasuwienas: You're right, I'll include it in this PR:
https://github.com/hashgraph/hedera-forking/pull/219

PR #215, opened 2025-01-22
fix: emit correct transfer and approval events (#194)

### LINKED ISSUE #194: Emit correct `Transfer` and `Approval` events for ERC721
The events declared for ERC721 in `IERC721Events` do not contain the `Transfer` and `Approval` methods. This is because the ERC721 implementation reuses the events from ERC20.

However, there is a slight difference. The `uint256` argument is `indexed` for ERC721, but not for ERC20. This can lead to inconsistencies.

Use and emit the proper event definitions for the ERC721 implementation.

arianejasuwienas:
**Description**:
Emit current transfer and approval events for ERC 721

**Related issue(s)**:

Fixes #194

**Notes for reviewer**:

**Checklist**


contracts/IERC20.sol
```diff
@@ -13,13 +13,13 @@ interface IERC20Events {
      *
      * Note that `value` may be zero.
      */
-    event Transfer(address indexed from, address indexed to, uint256 amount);
+    event ERC20Transfer(address indexed from, address indexed to, uint256 amount);
```
acuarica, line 16: we can't change the event signature (its name in this case) as this changes the ABI.

contracts/HtsSystemContract.sol
```diff
@@ -763,7 +763,7 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
 
         // Set the new owner
         assembly { sstore(slot, to) }
-        emit Transfer(from, to, serialId);
+        emit ERC721Transfer(from, to, serialId);
```
acuarica, line 766: We can change the contract definition to

```solidity
contract HtsSystemContract is IHederaTokenService {
```

and then emit the events like this

```suggestion
        emit IERC721Events.Transfer(from, to, serialId);
```

contracts/IERC721.sol
```diff
@@ -5,7 +5,9 @@ pragma solidity ^0.8.0;
  * See https://ethereum.org/en/developers/docs/standards/tokens/erc-721/#events for more information.
  */
 interface IERC721Events {
-    event ApprovalForAll(address indexed owner, address indexed operator, bool approved);
+    event ERC721Transfer(address indexed from, address indexed to, uint256 indexed tokenId);
+    event ERC721Approval(address indexed owner, address indexed approved, uint256 indexed tokenId);
+    event ERC721ApprovalForAll(address indexed owner, address indexed operator, bool approved);
```
acuarica, line 10: let's keep the same names to avoid changing the ABI and use `serialId` to be consistent with the rest of the implementation

```suggestion
    event Transfer(address indexed from, address indexed to, uint256 indexed serialId);
    event Approval(address indexed owner, address indexed approved, uint256 indexed serialId);
    event ApprovalForAll(address indexed owner, address indexed operator, bool approved);
```

acuarica, line 10: Also, take a look at the section https://github.com/hashgraph/hedera-forking?tab=readme-ov-file#non-fungible-tokens. These tables are auto-generated using the Foundry `IERC721` interface. Let's update the NatSpec docs on our `IERC721` so the docs come from it.

acuarica (changes requested):

contracts/IERC721.sol
acuarica: See at the end of section https://github.com/hashgraph/hedera-forking?tab=readme-ov-file#build on how to update the `README` from the interfaces.

arianejasuwienas: Won't be so easy. We can't have two events with the same selector defined in the same smart contract. This indexed flag is only an option for logger. Perhaps we should split the logic of the IERC20 and IERC721 into to separate contracts? I'll investigate the possibilities...

test/ERC721.t.sol
```diff
@@ -4,10 +4,9 @@ pragma solidity ^0.8.0;
 import {HtsSystemContract} from "../contracts/HtsSystemContract.sol";
 import {Test, console} from "forge-std/Test.sol";
 import {IERC721, IERC721Events} from "../contracts/IERC721.sol";
-import {IERC20Events} from "../contracts/IERC20.sol";
 import {TestSetup} from "./lib/TestSetup.sol";
 
-contract ERC721TokenTest is Test, TestSetup, IERC721Events, IERC20Events {
+contract ERC721TokenTest is Test, TestSetup {
```
acuarica, line 9: let's just implement `IERC721Events` so we don't have to change the `emit` here and also to be consistent with `Token.t.sol` (ERC20) test.

```suggestion
contract ERC721TokenTest is Test, TestSetup, IERC721Events {
```

contracts/IERC721.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity ^0.8.0;
  * See https://ethereum.org/en/developers/docs/standards/tokens/erc-721/#events for more information.
  */
 interface IERC721Events {
+    event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);
```
acuarica, line 8: as talked offline, let's include NatSpec docs here.

acuarica (changes requested):

andrewb1269hg (approved):
`README.md` file formatting looks good.

test/ERC721.t.sol
acuarica: let's get this done so we can merge this.

acuarica (approved):

arianejasuwienas:
@acuarica I need your approval again (lost it by resolving the conficts between this branch and main in HTS.t.sol file).

acuarica (approved):

andrewb1269hg (approved):
Reapprove readme formatting.

PR #216, opened 2025-01-22
feat: crypto transfer method implemented (#195)

### LINKED ISSUE #195: Providing an implementation for the cryptoTransfer method defined in the IHederaTokenService interface
The `cryptoTransfer` method from the IHederaTokenService interface is a functionality that enables transfers of both HBAR (Hedera’s native cryptocurrency) and tokens (fungible or non-fungible) within the Hedera network. 

It requires logic that is not yet present in the IERC721 and IERC20 interfaces, as it processes balance changes for fungible tokens in a different manner. Instead of handling individual transfers with sender, recipient, and amount details, it accounts for the net changes in balances (positive and/or negative) across all accounts. This approach requires ensuring that the overall balances reconcile to zero.

Additionally, it incorporates HBAR transfer functionality.

Interface to implement:
```
function cryptoTransfer(
    TransferList memory transferList, 
    TokenTransferList[] memory tokenTransfers
) payable htsCall external returns (int64 responseCode)
```

arianejasuwienas:
**Description**:
Adding crypto transfer method support to the hts. While calling the address 0x167 you the cryptoTransefer method can be invoked.

**Related issue(s)**:

Fixes #195

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -44,27 +44,8 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
     }
 
     function cryptoTransfer(TransferList memory transferList, TokenTransferList[] memory tokenTransfers)
-    payable htsCall external returns (int64 responseCode) {
-        uint256 hbarsReceived = msg.value;
-        int64 hbarBalance = 0;
-        for (uint256 hbarIndex = 0; hbarIndex < transferList.transfers.length; hbarIndex++) {
-            require(!transferList.transfers[hbarIndex].isApproval, "cryptoTransfer: hbar approval is not supported");
-            hbarBalance += transferList.transfers[hbarIndex].amount;
-            if (transferList.transfers[hbarIndex].amount < 0) {
-                require(
-                    transferList.transfers[hbarIndex].accountID == msg.sender,
-                    "cryptoTransfer: hbar transfer allowed only from the msg sender account"
-                );
-                continue;
-            }
-            require(transferList.transfers[hbarIndex].amount > 0, "cryptoTransfer: invalid amount");
-            uint256 value = uint256(uint64(transferList.transfers[hbarIndex].amount));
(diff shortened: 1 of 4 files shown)


contracts/HtsSystemContract.sol
```diff
@@ -43,6 +43,119 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         assembly { accountId := sload(slot) }
     }
 
+    function cryptoTransfer(TransferList memory transferList, TokenTransferList[] memory tokenTransfers)
+    payable htsCall external returns (int64 responseCode) {
```
arianejasuwienas, line 47: This method is not yet implemented in the services repository:

https://github.com/hashgraph/hedera-smart-contracts/blob/88472c9879fc02bc697b08e60fae11718d74d0b0/test/foundry/mocks/hts-precompile/HtsSystemContractMock.sol#L1359

so it does not necessarily have to be payable.

contracts/HtsSystemContract.sol
```diff
@@ -43,6 +43,119 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        uint256 hbarsReceived = msg.value;
```
acuarica, line 48: Not sure we are in the right track here. The wrapper method for `cryptoTransfer` https://github.com/hashgraph/hedera-smart-contracts/blob/0a617acfd804131e6f0ad5b7bc46019783f82bad/contracts/system-contracts/hedera-token-service/HederaTokenService.sol#L29-L35 does not pass `msg.value` so it does not seem this should be `payable`.

arianejasuwienas: It won't work in hardhat, but is accomplishable in foundry. I'll proceed with implementing that; I'm currently working on adding the approval mechanism.

Also, since I'm using vm deal cheatcodes to change balances now, I have to rethink / double check the validation and assertions.

PR #220, opened 2025-01-24
feat: [Foundry] implement `create*FungibleToken*`

### LINKED ISSUE #202: [Foundry] Implement token creation and deletion

### LINKED ISSUE #221: Fix `getTokenInfo`'s `totalSupply` after a `mintToken`/`burnToken`
After a call to `mintToken` or `burnToken`, the total supply of the token should change accordingly. However, when getting the token info with

```solidity
(int64 responseCode, IHederaTokenService.TokenInfo memory tokenInfo) = IHederaTokenService(HTS_ADDRESS).getTokenInfo(token);
```

the value `tokenInfo.totalSupply` does not reflect the change in the new total supply.

acuarica:
**Description**:

This PR implements the `create(Non)FungibleToken*` method family for Foundry. It uses the strategy outlined in `CREATE_TOKEN.md`. Essentially performing a temporarily deploy to set the `_tokenInfo` field.

The introduction of `_tokenInfo` duplicates the fields `name`, `symbol` and `totalSupply` (because these were top-level defined as well). ERC20 and ERC721 implementations were using the top-level fields. This implies that create token needs to write into two slots for the same value. Moreover, the `totalSupply` used by these interfaces was the top-level one as well. This introduces an issue were `getTokenInfo`'s `totalSupply` was not updated properly. See issue #221. This PR fixes this issue by removing the top-level fields and using the values from `_tokenInfo`. 

The `_initSlot` is introduced in favor of the `initialized` field. This allows us to easily compute the initialized slot in the `SetTokenInfo` contract.

Additionally, this PR also renames the ERC20 test to be consistent with the ERC721 test.

It also uses the `IHederaTokenService` in tests when possible. The code and interface of `HtsSystemContract` should be an implementation details. All accesses should be done through `IHederaTokenService`. The only accesses using `HtsSystemContract` directly should be internal testing, _e.g._, `getAccountId`.

**Related issue(s)**:

Fixes #202.
Fixes #221.

**Notes for reviewer**:

**Checklist**


contracts/HtsSystemContract.sol
```diff
@@ -6,18 +6,29 @@ import {IERC721, IERC721Events} from "./IERC721.sol";
...
+    // That is, they live in the token address storage space, not in the space of HTS `0x167`.
     // See `__redirectForToken` for more details.
-    string internal tokenType;
-    string internal name;
-    string internal symbol;
+    //
+    // Moreover, these variables must match the slots defined in `SetTokenInfo`.
+    string internal tokenType; 
+    // string internal name;
+    // string internal symbol;
```
quiet-node, line 29: qq: are these and the totalSupply meant to be commented out? if so why comment not remove just curious

contracts/HtsSystemContract.sol
```diff
@@ -125,6 +136,103 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        tokenInfo.fixedFees = fixedFees;
+        tokenInfo.fractionalFees = fractionalFees;
+        tokenInfo.royaltyFees = royaltyFees;
+        tokenInfo.ledgerId = "0x03";
+
+        bytes32 nextTokenIdSlot = _nextTokenIdSlot;
+        uint160 nextTokenId;
+        assembly { nextTokenId := sload(nextTokenIdSlot) }
+        if (nextTokenId == 0) {
+            nextTokenId = 1031;
```
quiet-node, line 178: This is a bit confusing to me. If `nextTokenId` is 0, it indicates `non-forking mode`. According to the design document, it states, “this counter should be 0.0.1032, the first token account ID created by the local node.” Shouldn’t this number be 1032? Or are we using zero-based indexing or some sort?

quiet-node: Ah, by the way, I know it’s mentioned in the doc, but I think adding a quick explanatory comment here to re-state why this number is used would be helpful. You know sometimes people dive into the code without reading the entire design document.

contracts/HtsSystemContract.sol
```diff
@@ -125,6 +136,103 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
...
+        }
+        nextTokenId++;
```
quiet-node, line 180: Ah, it’s incremented down here, so that would make it 1032? Ah, yeah, that makes sense.

quiet-node (dismissed):

contracts/HtsSystemContract.sol
acuarica: Indeed, nice catch!

contracts/HtsSystemContract.sol
acuarica: Exactly!

contracts/HtsSystemContract.sol
acuarica: I included a short description in the slot definition.

quiet-node (approved):
LGTM

PR #222, opened 2025-01-28
feat: returning correct ledger id in tokeninfo (#163)

### LINKED ISSUE #163: Fix `ledgerId` when forking from Local Node
The `ledgerId` does not return the proper value when forking from a local node network. This applies to both Foundry library and Hardhat plugin.

For the Foundry library, the solution is simpler because the `chainId` is already available. We just need to include another case for local node in `HtsSystemContractJson::_getLedgerId`.

On the other hand, for the Hardhat plugin, the core function `getHtsStorageAt` does not have JSON-RPC access. We could however extend the `IMirrorNodeClient` to support retrieving the chainId of the remote network.

arianejasuwienas:
**Description**:

The Ledger ID returned in TokenInfo will now provide the correct value.

1. In Hardhat, the Mirror Node URL will be used to determine the Ledger ID.
2. In Foundry, we already have access to the Chain ID, so we can use it to infer the Ledger ID.

**Related issue(s)**:

Fixes #163

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/src/forwarder/json-rpc-forwarder.js
+++ b/src/forwarder/json-rpc-forwarder.js
@@ -35,7 +35,7 @@ const {
 /** @type {Partial<import('hardhat/types').HardhatNetworkForkingConfig>} */
 const { url: forkUrl, mirrorNodeUrl, workerPort, localAddresses = [] } = workerData;
 
-require('../slotmap').init(workerData.chainId);
+require('../slotmap').setLedgerId(workerData.chainId);
 
 assert(mirrorNodeUrl !== undefined, 'json-rpc-forwarder: Missing Mirror Node URL');
 
--- a/src/forwarder/json-rpc-forwarder.js
+++ b/src/forwarder/json-rpc-forwarder.js
@@ -33,7 +33,9 @@ const {
 } = require('..');
 
 /** @type {Partial<import('hardhat/types').HardhatNetworkForkingConfig>} */
-const { url: forkUrl, mirrorNodeUrl, chainId, workerPort, localAddresses = [] } = workerData;
+const { url: forkUrl, mirrorNodeUrl, workerPort, localAddresses = [] } = workerData;
+
+require('../slotmap').init(workerData.chainId);
 
 assert(mirrorNodeUrl !== undefined, 'json-rpc-forwarder: Missing Mirror Node URL');
 
@@ -44,15 +46,7 @@ debug(
     mirrorNodeUrl
 );
 
-const chainIdToLedgerIdMap = {
-    295: '0x00',
-    296: '0x01',
-    297: '0x02',
-    298: '0x03',
-};
-const ledgerId =
-    chainIdToLedgerIdMap[/**@type{keyof typeof chainIdToLedgerIdMap}*/ (chainId)] || '0x00';
-const mirrorNodeClient = new MirrorNodeClient(mirrorNodeUrl, ledgerId);
+const mirrorNodeClient = new MirrorNodeClient(mirrorNodeUrl);
 
 /**
  * Function signature for `eth_*` method handlers.
--- a/src/forwarder/json-rpc-forwarder.js
+++ b/src/forwarder/json-rpc-forwarder.js
@@ -30,12 +30,13 @@ const {
     HTSAddress,
     LONG_ZERO_PREFIX,
     getHIP719Code,
+    setLedgerId,
 } = require('..');
 
 /** @type {Partial<import('hardhat/types').HardhatNetworkForkingConfig>} */
 const { url: forkUrl, mirrorNodeUrl, workerPort, localAddresses = [] } = workerData;
 
-require('../slotmap').setLedgerId(workerData.chainId);
+setLedgerId(workerData.chainId);
 
 assert(mirrorNodeUrl !== undefined, 'json-rpc-forwarder: Missing Mirror Node URL');
 
--- a/src/forwarder/json-rpc-forwarder.js
+++ b/src/forwarder/json-rpc-forwarder.js
@@ -33,7 +33,7 @@ const {
 } = require('..');
 
 /** @type {Partial<import('hardhat/types').HardhatNetworkForkingConfig>} */
-const { url: forkUrl, mirrorNodeUrl, workerPort, localAddresses = [] } = workerData;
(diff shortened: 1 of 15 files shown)


src/forwarder/mirror-node-client.js
```diff
@@ -48,6 +48,27 @@ class MirrorNodeClient {
         this.mirrorNodeUrl = mirrorNodeUrl;
     }
 
+    /**
+     * Get ledger ID.
+     *
+     * @returns {string}
+     */
+    getLedgerId() {
```
acuarica, line 56: this shouldn't depend on the `mirrorNodeUrl`, we need to devise another approach to get the `ledgerId`.

test/plugin/.testConfig.js
```diff
@@ -26,7 +26,7 @@ module.exports = {
      */
     projectTestConfig: {
         mocha: {
-            timeout: 30000,
+            timeout: 60000,
```
acuarica, line 29: restore this, I found out there were some tests failing due to timeouts as well. However, we need to see where this issue is coming from. By incrementing the timeout we are just masking the issue.

src/slotmap.js
```diff
@@ -24,6 +24,16 @@ const {
     storageLayout: { storage, types },
 } = require('../out/HtsSystemContract.sol/HtsSystemContract.json');
 
+let ledgerId = '0x00';
+/**
+ * @param {number|undefined} chainId
+ */
+const init = chainId => {
```
acuarica, line 31:
```suggestion
const setLedgerId = chainId => {
```

src/slotmap.js
```diff
@@ -24,6 +24,16 @@ const {
     storageLayout: { storage, types },
 } = require('../out/HtsSystemContract.sol/HtsSystemContract.json');
 
+let ledgerId = '0x00';
+/**
+ * @param {number|undefined} chainId
```
acuarica, line 29: where does `undefined` come from?

src/slotmap.js
```diff
@@ -24,6 +24,16 @@ const {
     storageLayout: { storage, types },
 } = require('../out/HtsSystemContract.sol/HtsSystemContract.json');
 
+let ledgerId = '0x00';
+/**
```
acuarica, line 28:
```suggestion
/**
 * Sets the `ledgerId` used when retrieving fungible and non-fungible tokens.
 * The `ledgerId` depends on the `chainId` of the remote network it is forking from.
```

src/slotmap.js
```diff
@@ -24,6 +24,16 @@ const {
     storageLayout: { storage, types },
 } = require('../out/HtsSystemContract.sol/HtsSystemContract.json');
 
+let ledgerId = '0x00';
```
acuarica, line 27: we should declared this as an `export` in `index.d.ts`.

acuarica (changes requested):

src/slotmap.js
arianejasuwienas: Do you mean the default value or the ledgerId variable itself? For now, it doesn’t seem like this is something we want to expose outside of our plugin. But if we do, perhaps it would be cleaner to move all of this logic to the index.js file instead?

acuarica: You're right, let's keep it as is.

src/slotmap.js
```diff
@@ -24,6 +24,19 @@ const {
...
+/**
+ * Sets the `ledgerId` used when retrieving fungible and non-fungible tokens.
+ * The `ledgerId` depends on the `chainId` of the remote network it is forking from.
+ *
+ * @param {number} chainId
+ */
+const setLedgerId = chainId => {
```
acuarica, line 34: let's export this one from `index.d.ts`. Also we can move the doc comments there, so consumers can take advantage of the js docs.

nit: convert to `function`.

src/forwarder/json-rpc-forwarder.js
```diff
@@ -28,6 +28,8 @@ const { getHtsCode, getHtsStorageAt, HTSAddress, LONG_ZERO_PREFIX, getHIP719Code
 /** @type {Partial<import('hardhat/types').HardhatNetworkForkingConfig>} */
 const { url: forkUrl, mirrorNodeUrl, workerPort, localAddresses = [] } = workerData;
 
+require('../slotmap').setLedgerId(workerData.chainId);
```
acuarica, line 31: import `setLedgerId` in the `require('..')` that is above.

```suggestion
setLedgerId(workerData.chainId);
```

We should treat exported modules in declated `package.json` as independent. All external used functions/classes/consts should be imported from an `index.js` declared in `package.json`. This way is easier to consumers to also use the package.

src/slotmap.js
```diff
@@ -339,4 +352,4 @@ class PersistentStorageMap {
     }
 }
 
-module.exports = { packValues, slotMapOf, PersistentStorageMap };
+module.exports = { packValues, slotMapOf, PersistentStorageMap, setLedgerId };
```
acuarica, line 355: also export `setLedgerId` through `index.js`.

acuarica (changes requested):

src/slotmap.js
arianejasuwienas: Okay, I think I understand the purpose now. I've exposed setLedgerId and added its description to the index.d.ts file. Let me know if it's okay now.

acuarica (commented):
Nice stuff!

Let's enable the test `"should retrieve token's metadata through getTokenInfo"` in `hts.e2e.js`.

test/hts.e2e.js
```diff
@@ -238,7 +238,7 @@ describe('::e2e', function () {
                 });
 
                 // To enable this, we need to change `getTokenInfo` to a `view` function
-                it.skip("should retrieve token's metadata through `getTokenInfo`", async function () {
+                it("should retrieve token's metadata through `getTokenInfo`", async function () {
```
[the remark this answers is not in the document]
acuarica: Sent a PR https://github.com/hashgraph/hedera-forking/pull/238 to fix both issues. Feel free to include the PR here.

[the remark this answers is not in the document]
arianejasuwienas: Merged it, thank you!

acuarica (approved):
Great work!

PR #223, opened 2025-01-29
fix: returning non-success response code instead of reverting when supplykey is missing (#167)

### LINKED ISSUE #167: Fix `mintToken/burnToken` when there is no supply key
If no supply key is set for an HTS token, then both minting and burning operations are disabled. They do not revert, but instead return code [`TOKEN_HAS_NO_SUPPLY_KEY`](https://github.com/hashgraph/hedera-services/blob/fb3467929f95f51547f2b7d2e8a51dffb52f4a15/hapi/hedera-protobufs/services/response_code.proto#L680)

```protobuf
    TOKEN_HAS_NO_SUPPLY_KEY = 180; // Supply key is not set on token
```

HTS emulation should mimic this behavior.

arianejasuwienas:
**Description**:

Mint and burn methods can not be executed when no supply key is set. But to make it work the same as on the mainnet we will return the response code 180 (HederaResponseCodes.TOKEN_HAS_NO_SUPPLY_KEY) then  instead of reverting the transaction.

**Related issue(s)**:

Fixes #167

**Notes for reviewer**:

**Checklist**


contracts/HtsSystemContract.sol
```diff
@@ -67,7 +67,9 @@ contract HtsSystemContract is IHederaTokenService, IERC20Events, IERC721Events {
         require(tokenInfoResponseCode == HederaResponseCodes.SUCCESS, "mintToken: failed to get token info");
 
         address treasuryAccount = tokenInfo.token.treasury;
-        require(treasuryAccount != address(0), "mintToken: invalid account");
+        if (treasuryAccount == address(0)) {
```
acuarica, line 70: This should be the supply key, not the treasury account.

acuarica (changes requested):
Looking good.

One more thing, there is a test in `hts.e2e.js` currently skipped `"should not mint because there is no supply key'"`. We should enable it with this change.

contracts/HtsSystemContract.sol
arianejasuwienas: OK, but this required us to support fetching the addresses of the keys associated with the tokens, which had a much larger scope than the fixed validation in this task. That's why I moved the most crucial changes to a separate PR: https://github.com/hashgraph/hedera-forking/pull/233

arianejasuwienas:
This is completly new approach to implement this task. That is why I've even removed all of the previous commits from the branch.

contracts/HtsSystemContract.sol
```diff
@@ -64,6 +72,9 @@ contract HtsSystemContract is IHederaTokenService {
         require(amount > 0, "mintToken: invalid amount");
 
         (int64 tokenInfoResponseCode, TokenInfo memory tokenInfo) = IHederaTokenService(token).getTokenInfo(token);
+        if (!ignoreSupplyKeyCheck && _extractKeyAddress(0x10, tokenInfo) == address(0)) { // 0x10 - supply key
```
acuarica, line 75: is this the right usage of `contractId`?

contracts/HtsSystemContract.sol
```diff
@@ -64,6 +72,9 @@ contract HtsSystemContract is IHederaTokenService {
...
+            return (HederaResponseCodes.TOKEN_HAS_NO_SUPPLY_KEY, tokenInfo.totalSupply, new int64[](0));
+        }
         require(tokenInfoResponseCode == HederaResponseCodes.SUCCESS, "mintToken: failed to get token info");
 
         address treasuryAccount = tokenInfo.token.treasury;
```
acuarica, line 80: I think the supply key account is where the tokens should be minted to.

acuarica (changes requested):
We need to include validation tests from now on to ensure the functionality we add or change matches real HTS.

contracts/HtsSystemContract.sol
arianejasuwienas: Treasury account receives that tokens, I've added the tests here:

https://github.com/hashgraph/hedera-forking/tree/167-correct-response-for-no-supply-key-analyse

And here are the results:
1. First I created a token, https://hashscan.io/testnet/token/0.0.5499059 with initial balance 10,000.00
. I've set  supplier to be: 
https://hashscan.io/testnet/account/0.0.4436613
2. Then as a supplier I minted 3 tokens.
3. As a supplier I still have 0 tokens of these type. The treasury account received all of them: https://hashscan.io/testnet/account/0.0.5416580

contracts/HtsSystemContract.sol
arianejasuwienas: Sadly no, it was more of a hack .

acuarica:
> That is why I've even removed all of the previous commits from the branch.

No need, better to keep commits for future reference so others can see the what has been done.

contracts/HtsSystemContract.sol
acuarica: Please include it in the same PR, otherwise it's difficult to comment on.

acuarica: > Treasury account receives that tokens

Yes, got confused. Thanks for the clarification.

contracts/HtsSystemContract.sol
```diff
@@ -940,4 +954,14 @@ contract HtsSystemContract is IHederaTokenService {
         assembly { sstore(slot, approved) }
         emit IERC721.ApprovalForAll(sender, operator, approved);
     }
+
+    function _keyExists(uint keyType, TokenInfo memory tokenInfo) private pure returns (bool) {
```
acuarica, line 958: let's move this into its own `library` so we can put all key related stuff in there.

name: `KeyLib`? `KeyMgmt`? other alternatives?

contracts/HtsSystemContract.sol
```diff
@@ -940,4 +954,14 @@ contract HtsSystemContract is IHederaTokenService {
...
+        for (uint256 i = 0; i < tokenInfo.token.tokenKeys.length; i++) {
+            if (tokenInfo.token.tokenKeys[i].keyType == keyType) {
```
acuarica, line 960: `keyType` is a bit field, so we need to check if the right bit is set. However, we can leave it for another PR.

test/HTS.t.sol
```diff
@@ -239,6 +242,33 @@ contract HTSTest is Test, TestSetup {
         IHederaTokenService(HTS_ADDRESS).mintToken(token, amount, metadata);
     }
 
+    function test_mintToken_should_fail_with_np_supplier() external {
```
acuarica, line 245:
```suggestion
    function test_mintToken_should_fail_with_no_supplyKey() external {
```

test/HTS.t.sol
```diff
@@ -218,6 +218,9 @@ contract HTSTest is Test, TestSetup {
         bytes[] memory metadata = new bytes[](0);
         int64 initialTotalSupply = 10000000005000000;
         IHederaTokenService.TokenInfo memory tokenInfo;
+        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](1);
+        keys[0].keyType = 0x10; // Supply key
+        keys[0].key.contractId = msg.sender;
```
acuarica, line 223: let's set some ECDSA key to test this instead, so if we change the `contractId` semantics this test still passes.

acuarica (changes requested):
Also please enable `should not mint because there is no supply key' in `hts.e2e.js` test. I checked and the test now passes. After that we should be good to go.

contracts/HtsSystemContract.sol
acuarica: No need to include the branch contents.

contracts/HtsSystemContract.sol
```diff
@@ -56,11 +57,22 @@ contract HtsSystemContract is IHederaTokenService {
         int64 responseCode,
         int64 newTotalSupply,
         int64[] memory serialNumbers
+    ) {
+        return _mintToken(token, amount, true);
+    }
+
+    function _mintToken(address token, int64 amount, bool checkSupplyKey) htsCall internal returns (
```
acuarica, line 64:
```suggestion
    function _mintToken(address token, int64 amount, bool checkSupplyKey) private returns (
```

acuarica (approved):
Great stuff! Thanks for the effort!

PR #225, opened 2025-01-30
feat: getnftinfo will return its metadata and creation time

### LINKED ISSUE #198: NFT metadata and creation time
The IHederaTokenService interface requires non-fungible token (NFT) information to include metadata and creation time, which are not part of the IERC721 interface and are currently unavailable for retrieval. To support this functionality, it is necessary to store the creation time in the token’s storage and use base64 encoding to generate metadata from the tokenUri (which is defined in the IERC721 interface).

arianejasuwienas:
**Description**:

A call to the getNonFungibleTokenInfo method on the HTS address (0x167) will now correctly populate the "metadata" and "creationTime" fields.

**Related issue(s)**:

Fixes #198

**Notes for reviewer**:

Until now, these fields were mocked because they were not supported by IERC721 and required a direct implementation.

**Checklist**

### DIFF
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -436,18 +436,10 @@ contract HtsSystemContract is IHederaTokenService {
     function getNonFungibleTokenInfo(address token, int64 serialNumber)
         htsCall external view
         returns (int64, NonFungibleTokenInfo memory) {
-        (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
-        require(responseCode == HederaResponseCodes.SUCCESS, "getNonFungibleTokenInfo: failed to get token data");
-        (, NonFungibleTokenInfo memory nonFungibleTokenInfo) = IHederaTokenService(token).getNonFungibleTokenInfo(
+        return IHederaTokenService(token).getNonFungibleTokenInfo(
             token,
             serialNumber
         );
-        nonFungibleTokenInfo.tokenInfo = tokenInfo;
-        nonFungibleTokenInfo.serialNumber = serialNumber;
-        nonFungibleTokenInfo.spenderId = IERC721(token).getApproved(uint256(uint64(serialNumber)));
-        nonFungibleTokenInfo.ownerId = IERC721(token).ownerOf(uint256(uint64(serialNumber)));
-
-        return (responseCode, nonFungibleTokenInfo);
     }
 
     function isToken(address token) htsCall external view returns (int64, bool) {
@@ -627,11 +619,16 @@ contract HtsSystemContract is IHederaTokenService {
             if (selector == this.getNonFungibleTokenInfo.selector) {
                 require(msg.data.length >= 92, "getNonFungibleTokenInfo: Not enough calldata");
                 uint256 serialId = uint256(bytes32(msg.data[60:92]));
-                NonFungibleTokenInfo memory info;
+                NonFungibleTokenInfo memory nonFungibleTokenInfo;
                 (int64 creationTime, bytes memory metadata) = __nftInfo(serialId);
-                info.creationTime = creationTime;
-                info.metadata = metadata;
-                return abi.encode(HederaResponseCodes.SUCCESS, info);
+                nonFungibleTokenInfo.tokenInfo = _tokenInfo;
+                nonFungibleTokenInfo.serialNumber = int64(int256(serialId));
+                nonFungibleTokenInfo.ownerId = __ownerOf(serialId);
+                nonFungibleTokenInfo.creationTime = creationTime;
+                nonFungibleTokenInfo.metadata = metadata;
+                nonFungibleTokenInfo.spenderId = __getApproved(serialId);
+
+                return abi.encode(HederaResponseCodes.SUCCESS, nonFungibleTokenInfo);
             }
             if (selector == this._update.selector) {
                 require(msg.data.length >= 124, "update: Not enough calldata");
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -438,16 +438,15 @@ contract HtsSystemContract is IHederaTokenService {
         returns (int64, NonFungibleTokenInfo memory) {
         (int64 responseCode, TokenInfo memory tokenInfo) = getTokenInfo(token);
         require(responseCode == HederaResponseCodes.SUCCESS, "getNonFungibleTokenInfo: failed to get token data");
-        NonFungibleTokenInfo memory nonFungibleTokenInfo;
+        (, NonFungibleTokenInfo memory nonFungibleTokenInfo) = IHederaTokenService(token).getNonFungibleTokenInfo(
+            token,
+            serialNumber
+        );
         nonFungibleTokenInfo.tokenInfo = tokenInfo;
         nonFungibleTokenInfo.serialNumber = serialNumber;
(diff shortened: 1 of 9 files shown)


contracts/MirrorNode.sol
```diff
@@ -40,6 +40,14 @@ abstract contract MirrorNode {
         return "";
     }
 
+    function getNftCreatedTimestamp(address token, uint32 serial) external returns (string memory) {
+        string memory json = this.fetchNonFungibleToken(token, serial);
```
acuarica, line 44: let's rename `fetchNonFungibleToken` to `fetchNonFungibleTokenSerial`. This is to avoid confusion between the NFT and the actual token (which includes the serial). Also to be consistent with the rest.

contracts/HtsSystemContractJson.sol
```diff
@@ -463,6 +463,17 @@ contract HtsSystemContractJson is HtsSystemContract {
         return slot;
     }
 
+    function __nftInfoSlot(uint32 serialId) internal override virtual returns (bytes32) {
+        bytes32 slot = super.__nftInfoSlot(serialId);
+        if (_shouldFetch(slot)) {
+            string memory metadata = mirrorNode().getNftMetadata(address(this), serialId);
+            string memory createdTimestamp = mirrorNode().getNftCreatedTimestamp(address(this), serialId);
```
acuarica, line 470: we should unify these two calls to the MN into one, maybe something like `getNftMetadataAndCreatedTimestamp`, and extract both fields.

contracts/HtsSystemContractJson.sol
```diff
@@ -463,6 +463,17 @@ contract HtsSystemContractJson is HtsSystemContract {
...
+            int64 creationTime = int64(vm.parseInt(vm.split(createdTimestamp, ".")[0]));
+            storeBytes(address(this), uint256(slot), abi.encode(creationTime, bytes(metadata)));
```
acuarica, line 472: we should also set the flag in `_scratchAddr()` to prevent re-fetching (see `_setValue` function).

Also, please modify `_tokenUriSlot` because it doesn't update the `_scratchAddr()` either.

contracts/HtsSystemContract.sol
```diff
@@ -593,6 +584,19 @@ contract HtsSystemContract is IHederaTokenService {
                 _setApprovalForAll(from, to, approved);
                 return abi.encode(true);
             }
+            if (selector == this.getNonFungibleTokenInfo.selector) {
+                require(msg.data.length >= 92, "getNonFungibleTokenInfo: Not enough calldata");
+                uint256 serialId = uint256(bytes32(msg.data[60:92]));
+                NonFungibleTokenInfo memory nonFungibleTokenInfo;
+                (int64 creationTime, bytes memory metadata) = __nftInfo(serialId);
+                nonFungibleTokenInfo.creationTime = creationTime;
```
acuarica, line 592: nit: let's use the same ordering as in `NonFungibleTokenInfo` so it's easier to see that all fields are being set.

src/index.js
```diff
@@ -244,7 +244,40 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
                 ZERO_HEX_32_BYTE,
                 `Failed to get the metadata of the NFT ${tokenId}#${serialId}`
             );
-        persistentStorage.store(tokenId, blockNumber, nrequestedSlot, atob(metadata));
+        persistentStorage.store(
+            tokenId,
+            blockNumber,
+            nrequestedSlot,
+            atob(metadata),
```
acuarica, line 251: Let's use `Buffer.from(metadata, 'base64').toString()` to be consistent with similar code below.

src/index.js
```diff
@@ -244,7 +244,40 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
...
+            metadata: null,
+            created_timestamp: null,
+        };
+        if (typeof metadata !== 'string' || typeof created_timestamp !== 'string')
+            return ret(
+                ZERO_HEX_32_BYTE,
+                `Failed to get the metadata of the NFT ${tokenId}#${serialId}`
+            );
+        const timestamp = Number(created_timestamp.split('.')[0]).toString(16).padStart(64, '0');
+        const metadataEncoded = Buffer.from(metadata).toString('hex');
```
acuarica, line 276: did you mean

```suggestion
        const metadataEncoded = Buffer.from(metadata, 'base64').toString('hex');
```
?

src/slotmap.js
```diff
@@ -316,10 +316,11 @@ class PersistentStorageMap {
      * @param {number} blockNumber
      * @param {bigint} slot
      * @param {Value} value
+     * @param {string} type
```
acuarica, line 319: for now let's use a more restrictive type

```suggestion
     * @param {'t_string_storage' | 't_bytes_storage'} type
```

acuarica (changes requested):
Looking good, thanks for sending this.

src/index.js
arianejasuwienas: No, the metadata is just a raw, encoded value - it can be any string. I'm simply calculating the bytes of the string without doing anything else to it. @acuarica

arianejasuwienas:
@acuarica Thank you for review! I've applied your suggestions (except for this one: https://github.com/hashgraph/hedera-forking/pull/225#discussion_r1967424830, where I've left a comment).

src/index.js
```diff
@@ -260,7 +260,40 @@ async function getHtsStorageAt(address, requestedSlot, blockNumber, mirrorNodeCl
...
+        if (typeof metadata !== 'string' || typeof created_timestamp !== 'string')
+            return ret(
+                ZERO_HEX_32_BYTE,
+                `Failed to get the metadata of the NFT ${tokenId}#${serialId}`
+            );
+        const timestamp = Number(created_timestamp.split('.')[0]).toString(16).padStart(64, '0');
+        const metadataEncoded = Buffer.from(metadata, 'utf-8').toString('hex');
+        const metadataLength = metadata.length.toString(16).padStart(64, '0');
+        const stringTypeIndicator = '40'.padStart(64, '0');
+        const bytes = `${timestamp}${stringTypeIndicator}${metadataLength}${metadataEncoded}`;
```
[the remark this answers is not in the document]
arianejasuwienas: Below are the contents of the memory starting at this slot:

https://github.com/hashgraph/hedera-forking/pull/225/files#diff-6b0194add020b3f0fb2f67f2eac675de98278882a61d442e1c7a2b9bd97b41d4R517

It is a tuple containing an int64 followed by a string, which means it may occupy more than one slot. Note that IERC721 does not use it; instead, tokenURI is stored in a separate slot. I haven't changed that for now because it makes sense - we avoid having to decode the metadata value every time tokenURI is queried. Let me know if we want to change that.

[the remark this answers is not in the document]
arianejasuwienas: Here is this implementation with dedicated tokenUri slot removed:

https://github.com/hashgraph/hedera-forking/pull/250

One test would need to be rewritten, since we have no byte comparating function prepared yet to handle this operation; since I'm not sure if we want to go this way I've skipped this test instead of rewriting it for now.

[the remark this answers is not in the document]
arianejasuwienas: I've merged this (#250), so all changes are already included in this PR. @acuarica

acuarica:
@arianejasuwienas marked this as draft because moving forward we should also include validation tests.

PR #229, opened 2025-02-05
feat: export Solidity interfaces

### LINKED ISSUE #227: Export Solidity interfaces for Hardhat and Foundry
The HTS emulation contract consumes multiple interfaces to ensure adherence to different standards, `IERC20`, `IERC721`, `IHRC719` and other interfaces might be added in the future. Also `IHederaTokenService` is used. However, the user consuming the library/plugin needs to duplicate this interfaces in their projects.

Export said interfaces so that users can reuse these definitions.

acuarica:
**Description**:

This PR exports the necessary interfaces so that the library/plugin is self-contained. This allows the user to use the interfaces we provide.

In addition, it merges ERC20/ERC721 event and function interfaces. This allows for a better DX, given the user would need to import only one interface instead of two (for example `IERC20` and `IERC20Events`).

**Related issue(s)**:

Fixes #227.

**Notes for reviewer**:

**Checklist**


quiet-node (approved):
LGTM

PR #233, opened 2025-02-06
feat: fetching the addresses of the token key owners (#167)

### LINKED ISSUE #167: Fix `mintToken/burnToken` when there is no supply key

arianejasuwienas:
**Description**:

To accurately identify the token supplier or perform any operations dependent on these keys, we first needed to add support for them by determining their EVM addresses

**Related issue(s)**:

Fixes #167

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/HtsSystemContract.sol
+++ b/contracts/HtsSystemContract.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 import {IERC20} from "./IERC20.sol";
 import {IERC721} from "./IERC721.sol";
+import {ITokenKey} from "./ITokenKey.sol";
 import {IHRC719} from "./IHRC719.sol";
 import {IHederaTokenService} from "./IHederaTokenService.sol";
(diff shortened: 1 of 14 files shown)


arianejasuwienas:
After analyzing how this was implemented in the Hedera smart contracts (which rely entirely on the contractId address field), I believe it could be done much more simply. For this reason, this approach has been abandoned, and I'm closing this PR.

PR #234, opened 2025-02-07
feat: implement token creation for Hardhat

### LINKED ISSUE #203: [Hardhat] Implement token creation and deletion

acuarica:
**Description**:

This PR implements token creation for Hardhat.

It creates tokens locally with one deployment (instead of two as before) because `eth_getCode` is called only one by Hardhat and then cached. So there was no possibility of changing contract bytecode once it was deployed.

**Related issue(s)**:

Fixes #203.

**Notes for reviewer**:

**Checklist**


quiet-node (approved):
LGTM! Nice docs as always!

Issue #197, opened 2025-01-15
Implement Token Freeze, KYC, and Pause Functionalities


arianejasuwienas:
The `IHederaTokenService` interface provides support for functionalities beyond those defined in the IERC20 and IERC721 interfaces, including token freezing, KYC, and pausing features. These capabilities need to be implemented to ensure full support for default HTS methods.

PR #236, opened 2025-02-10
feat: token freeze, kyc and pause (#197)

arianejasuwienas:
**Description**:
Adds the support for token freeze, kyc and pause

CryptoTransfer and Key Support included (both functionalities are on different PullRequests, so this one should be merged last).

**Related issue(s)**:

#197

**Notes for reviewer**:

**Checklist**


arianejasuwienas:
@acuarica 

The key check is implemented, but it only verifies the key's existence. To determine whether the call can be made, we need to extract the key owner.

If you prefer, we can create a separate issue for the validation. Let me know what works best for you.

arianejasuwienas:
@acuarica I'm closing this PR because it verifies the key owner, but the implementation includes a lot of code. Instead, I've created a new PR where the key owner is not verified (only key existence is checked) to simplify the review process (as you suggested): https://github.com/hashgraph/hedera-forking/pull/256

PR #243, opened 2025-02-14
feat: isassociated method will revert for non-existing accounts (#242)

### LINKED ISSUE #242: Method `isAssociated` should revert for non-existing accounts
### Description

Our current implementation of the isAssociated method returns false when querying for a non-existent account. This behavior differs from that of the real, non-emulated HTS, where revert is performed in tis scenario.

### Steps to reproduce

This can be observed when running the:

```
should get not `isAssociated` for non-existing non-associated accounts
```
test from the `hts.e2e` test set.

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.1.1

### Operating system

None

arianejasuwienas:
**Description**:

This will make isAssociated method call for non-existing account revert, the same as it does on the real HTS.

**Related issue(s)**:

Fixes #242

**Notes for reviewer**:

**Checklist**


arianejasuwienas:
@acuarica This version utilizes the same slot in which we are storing getAccountId instead of a separate one (using the value of its first byte). Thanks to this, we can set both values simultaneously rather than querying them individually. JSON-RPC API usage in forking mode will be limited to a single call to check the initial value of the field which will return not only account id but also the information if it exists or not.

Take a look at this commit to see what changes were needed:

https://github.com/hashgraph/hedera-forking/pull/243/commits

test/hts.e2e.js
```diff
@@ -225,13 +225,17 @@ describe('::e2e', function () {
...
+                it('should get not `isAssociated` for non-existing non-associated accounts', async function () {
+                    const fails = async function (/** @type {string} */ from) {
+                        try {
+                            await ERC20['isAssociated']({ from });
+                        } catch {
+                            return true;
+                        }
+                        return false;
+                    };
+                    expect(await fails(nonAssocAddress0)).to.be.true;
```
acuarica, line 237: better to use `expect(() => ...).to.throw()` for clarity.

contracts/HtsSystemContract.sol
```diff
@@ -772,6 +769,7 @@ contract HtsSystemContract is IHederaTokenService {
             return abi.encode(true);
         }
         if (selector == IHRC719.isAssociated.selector) {
+            require(IHederaAccounts(HTS_ADDRESS).accountExists(msg.sender));
             bytes32 slot = _isAssociatedSlot(msg.sender);
```
acuarica, line 773: I think the change should be in `_isAssociatedSlot`.

contracts/HtsSystemContract.sol
```diff
@@ -845,22 +843,22 @@ contract HtsSystemContract is IHederaTokenService {
...
+        uint32 ownerId = IHederaAccounts(HTS_ADDRESS).getAccountId(owner);
+        uint32 spenderId = IHederaAccounts(HTS_ADDRESS).getAccountId(spender);
         return bytes32(abi.encodePacked(selector, pad, spenderId, ownerId));
     }
 
     function _isAssociatedSlot(address account) internal virtual returns (bytes32) {
         bytes4 selector = IHRC719.isAssociated.selector;
         uint192 pad = 0x0;
-        uint32 accountId = HtsSystemContract(HTS_ADDRESS).getAccountId(account);
+        uint32 accountId = IHederaAccounts(HTS_ADDRESS).getAccountId(account);
```
acuarica, line 861: I think we should test if `accountId` exists here, by checking the exists byte in `accountId`.

arianejasuwienas: For now, getAccountId was returning four bytes (uint32) and was always used as such. I attempted to rewrite it in the proposed way, but while writing the method description:

```
    /// @dev Returns the result in a format where:
    /// - The first byte indicates whether the account exists on the network we are forking from,
    ///   as this may influence the behavior of the true Hedera Token Service, so we replicate it here.
    /// - The last four bytes store a `uint32`, which is the account ID (excluding both shard and realm numbers)
    ///   for the given `address`. The storage adapter, _i.e._, `getHtsStorageAt`, assumes that both shard and realm
    ///   numbers are zero, allowing them to be omitted from the account ID.
    ///
    /// See https://docs.hedera.com/hedera/core-concepts/accounts/account-properties
    /// for more info on account properties.
 ```
I realized that a much cleaner and simpler approach would be to use a tuple in the method’s interface, where the first element represents the account ID and the second is a flag indicating whether the account exists - the true result we will actually want to use.

File: test/hts.e2e.js (hunk 1 above)
Comment (arianejasuwienas) [reply test/hts.e2e.js]:
```
expect(() => ...).to.throw()
```
Won't work for async functions, even with `eventually`. So, I used asserts instead—I think it's clear enough this way too,let me know.
@acuarica

File: contracts/HtsSystemContract.sol (hunk 3 above)
Comment (acuarica) [reply contracts/HtsSystemContract.sol]: Totally, that's the way to go.

**Review by acuarica — COMMENTED** [review]
Please merge PR mentioned in comments to simplify redirectForToken.

File: contracts/HtsSystemContract.sol [hunk 4]
```diff
@@ -845,22 +843,23 @@ contract HtsSystemContract is IHederaTokenService {
     function _balanceOfSlot(address account) internal virtual returns (bytes32) {
         bytes4 selector = IERC20.balanceOf.selector;
         uint192 pad = 0x0;
-        uint32 accountId = HtsSystemContract(HTS_ADDRESS).getAccountId(account);
+        (uint32 accountId, ) = IHederaAccounts(HTS_ADDRESS).getAccountId(account);
```
acuarica, line 846: If we use `getAccountId` directly from `HtsSystemContract`, we don't need `IHederaAccounts`.

contracts/MirrorNode.sol
```diff
@@ -124,6 +124,13 @@ abstract contract MirrorNode {
         return address(uint160(accountNum));
     }
 
+    function accountExist(address account) external returns (bool) {
```
acuarica, line 127: I inlined this into `HtsSystemContractJson`, no use case (at least for now) to have it here.

test/data/getAccount_0x0000000000000000000000000000000000001438.json
```diff
@@ -0,0 +1,36 @@
+{
```
acuarica, line 1: Made this a symlink (to the already existing account) to be explicit about what was added and avoid duplication.

test/hts.e2e.js
```diff
@@ -225,12 +225,14 @@ describe('::e2e', function () {
 
                 // These reverts instead of returning `false`.
                 // Should we have the same behavior in emulated HTS?
-                it.skip('should get not `isAssociated` for non-existing non-associated accounts', async function () {
-                    expect(await ERC20['isAssociated']({ from: nonAssocAddress0 })).to.be.equal(
-                        false
+                it('should get not `isAssociated` for non-existing non-associated accounts', async function () {
+                    await assert.rejects(
+                        () => ERC20['isAssociated']({ from: nonAssocAddress0 }),
+                        Error
```
acuarica, line 231: Nice alternative!

Can we find a more specific `Error` or error message to check here?

acuarica (approved):
Great stuff!

test/hts.e2e.js
acuarica: Thanks for the clarification.

test/hts.e2e.js
arianejasuwienas: The problem is that the error messages are actually different but it's because different json rpc api behaviour.
Local node returns data=null on this call, while anvil/local-node returns data="0x" (because that what it does with an empty data).

This causes the main client error message to change from:
```
missing revert data
```
to
```
execution reverted (no data present; likely require(false) occurred 
```
Let me know if we want to do anything about this. Maybe adding a comment about the behavioral differences would be sufficient? I'm not sure if this comment belongs in this specific test, though, since it seems more like a global issue.

acuarica: Good to know, thanks for the clarification. Let's leave like this for now. Maybe we can think of something for the future.

PR #249, opened 2025-02-24
chore: add solidity coverage in ci

### LINKED ISSUE #124: Add Solidity coverage in CI
Foundry supports contract coverage out of the box using `forge coverage`[^1].

Add a new CI job (or modify a existing one) to run contract coverage.

It would be also helpful to include the coverage summary as a PR comment.

[^1]: https://book.getfoundry.sh/reference/forge/forge-coverage

arianejasuwienas:
**Description**:

Adds Solidity Coverage in CI, publishes it in GitHub Comment.

**Related issue(s)**:

Fixes #124

**Notes for reviewer**:

**Checklist**


acuarica (commented):
Great start! Try to take a look at https://github.com/fingerprintjs/action-coverage-report-md. To simplify the reporting. Also, in case you needed, `forge coverage` supports `lcov` reporting with `--report lcov`, it might be helpful to have coverage in this format.

arianejasuwienas:
@acuarica Ok, I followed your suggestion. I used a third-party solution instead of a custom curl and switched to lcov report instead of summary. However, I'm not sure if this is the best approach because:

- [fingerprintjs/action-coverage-report-md](https://github.com/fingerprintjs/action-coverage-report-md) requires the Istanbul text format, not lcov (so I had to write a converter).
- The lcov format does not count statements on its own, so I used lines as the closest approximation.

.github/workflows/test.yml
```diff
@@ -232,3 +232,56 @@ jobs:
...
+          forge build
+
+      - name: Generate Coverage Report
+        run: |
+          forge coverage --report lcov
+          cat lcov.info
+
+      - name: Convert LCOV to Istanbul Text Format
+        run: |
+          .github/lcovtoistanbul lcov.info > istanbul.txt
```
acuarica, line 274: instead of manually converting the report, shouldn't we use a package to do this? See dependencies of  https://github.com/bcoe/c8, we might find something useful. I mentioned `c8` because that's what we will usef or JS coverage.

acuarica (commented):
Great start.

acuarica (changes requested):
Let's keep it simple for now. We can just run `forge coverage` in a job similar to `sol-tests-local-storage`. We might revisit this is in the future to comment on the PR.

And let's simplify the `sol-tests-local-storage` job. I see the step to setup Node, but it shouldn't be needed. If that's the case, remove the step.

Thanks for the effort!

acuarica (approved):

rbarker-dev (approved):

PR #251, opened 2025-02-25
chore: support for windows os

### LINKED ISSUE #153: [Foundry] Include Windows support in Foundry library
The Foundry library assumes it is running on a Unix-like operating system. It uses `curl` (`tail`, `sed`, `tr` and `bash`) to make requests and parse responses.

Windows users would need to install these tools separately in order to use the Foundry library.

Add Windows support by using commands that are available in Windows out-of-the-box.

### Alternative

Add the Unix-like tools as prerequisites to use the library. Users would need to install a Linux subsystem in order to use it.

arianejasuwienas:
**Description**:

Windows OS will be supported.

**Related issue(s)**:

Fixes #153

**Notes for reviewer**:

Need to test it on my windowsos, making it draft for now.

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ ffi = true
 
 Alternatively, you can add the `--ffi` flag to your execution script.
 
-This is necessary because:
+This is necessary because
 
 - On Unix-like operating systems, the library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network. This allows the library to fetch token state from the remote network. Since `curl` is an external command, `ffi` must be enabled.
 - On Windows, if PowerShell is available, the library uses the `Invoke-WebRequest` command to handle HTTP requests, which also requires `ffi` to be enabled.
@@ -66,7 +66,7 @@ This is necessary because:
 > The Foundry library assumes it is running on either a Windows system with PowerShell or a Unix-like operating system.
 >
 > - On Unix-like systems, it uses `curl`, `bash`, `tail`, `sed`, and `tr` to make requests and parse responses.
-> - On Windows, it uses `Invoke-WebRequest` and `Start-Process` in PowerShell for similar operations.
+> - On Windows, it uses `Invoke-WebRequest` and `Start-Process` in PowerShell for a similar effect.
 
 To activate HTS emulation in your tests, you need to add the following setup code in your test files.
 Import our wrapper function to deploy HTS emulation and enable cheat codes for it.
--- a/README.md
+++ b/README.md
@@ -57,13 +57,16 @@ ffi = true
 
 Alternatively, you can add the `--ffi` flag to your execution script.
 
-This is necessary because our library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network.
-This enables the library to fetch token state in the remote network.
-Given `curl` is an external command, `ffi` needs to be enabled.
+This is necessary because:
 
-> [!NOTE]
-> The Foundry library assumes it is running on a Unix-like operating system.
-> In addition to `curl`, it uses `bash`, `tail`, `sed` and `tr` and to make requests and parse responses.
+- On Unix-like operating systems, the library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network. This allows the library to fetch token state from the remote network. Since `curl` is an external command, `ffi` must be enabled.
+- On Windows, if PowerShell is available, the library uses the `Invoke-WebRequest` command to handle HTTP requests, which also requires `ffi` to be enabled.
+
+> [!NOTE]  
+> The Foundry library assumes it is running on either a Windows system with PowerShell or a Unix-like operating system.
+>
+> - On Unix-like systems, it uses `curl`, `bash`, `tail`, `sed`, and `tr` to make requests and parse responses.
+> - On Windows, it uses `Invoke-WebRequest` and `Start-Process` in PowerShell for similar operations.
 
 To activate HTS emulation in your tests, you need to add the following setup code in your test files.
 Import our wrapper function to deploy HTS emulation and enable cheat codes for it.
@@ -581,6 +584,10 @@ In case needed, there is a trace log of requests made by mocked `curl` in `scrip
 ```console
 cat test/scripts/curl.log
 ```
+
+> [!NOTE]
+> Since the Windows-based solution does not use `curl`, the mocked `curl` behavior will have no effect on it.
+> Mocked `curl` is designed exclusively for Unix-like operating systems.
 
 ### `HtsSystemContract` Solidity tests + JSON-RPC mock server for storage emulation (_with_ forking)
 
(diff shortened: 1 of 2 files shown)


contracts/Surl.sol
```diff
@@ -42,4 +48,91 @@ library Surl {
...
+        bytes memory res = vm.ffi(inputs);
+        (status, data) = abi.decode(res, (uint256, bytes));
+    }
+
+    function _cmd(string memory url) internal returns (uint256 status, bytes memory data) {
+        string[] memory inputs = new string[](3);
+        inputs[0] = "bash";
+        inputs[1] = "-c";
+        inputs[2] = string.concat(
+            "powershell -Command \"$response = Invoke-WebRequest -Uri '",
```
acuarica, line 70: which alternative is more ubiquitous? just for reference, which one is present where?

contracts/Surl.sol
```diff
@@ -42,4 +48,91 @@ library Surl {
...
+    }
+
+    function _isWindowsOS() internal returns (bool) {
+        string[] memory inputs = new string[](3);
+        inputs[0] = "cmd";
+        inputs[1] = "/C";
+        inputs[2] = "echo %OS%";
+
+        bytes memory res;
+        try vm.ffi(inputs) returns (bytes memory output) {
```
acuarica, line 85: couldn't we assume that if `cmd` succeeds then OS is Windows?

contracts/Surl.sol
```diff
@@ -42,4 +48,91 @@ library Surl {
...
+        inputs[2] = "ver";
+        try vm.ffi(inputs) returns (bytes memory output) {
+            res = output;
+            string memory osVersion = string(res);
+            return _containsSubstring(osVersion, "Windows ");
+        } catch {}
+        return false;
+    }
+
+    function _isPowerShellAvailable() internal returns (bool available) {
```
acuarica, line 99: See comment about alternatives.

acuarica (changes requested):
Great start, some stuff we can iterate on.

contracts/Surl.sol
```diff
@@ -42,4 +48,91 @@ library Surl {
...
+        inputs[2] = string.concat(
+            "powershell -Command \"$response = Invoke-WebRequest -Uri '",
+            url,
+            "' -UseBasicParsing; $status = $response.StatusCode; $body = $response.Content -join ''; Write-Output \"$status`n$body\"\""
+        );
+        bytes memory res = vm.ffi(inputs);
+        (status, data) = abi.decode(res, (uint256, bytes));
+    }
+
+    function _isWindowsOS() internal returns (bool) {
```
acuarica, line 78: or is there an environment variable to test the kind of OS? I saw there is `OSTYPE`in unix-like, but not sure on windows though.

contracts/Surl.sol
arianejasuwienas: Older versions of Windows didn’t have PowerShell. The solution I was working on required creating and using a temporary file then, but Windows Defender flagged this as suspicious and claimed my software was a virus :/. In practice, it’s unlikely anyone is using a Windows version without PowerShell today, so I eventually just removed support for them. Let me know if we can stick with the current solution. @acuarica

contracts/Surl.sol
arianejasuwienas: I'm doing it the other way around now - assuming that if uname is present, we most likely have a Unix-based system.

arianejasuwienas:
@acuarica Perhaps it would be helpful to mention Windows issues with symlinks in the README, since we are using them. However, I also believe that any developer working on Windows should be aware of these limitations and know their workarounds. Let me know what you think.

contracts/Surl.sol
```diff
@@ -29,7 +29,12 @@ library Surl {
      * @return status The HTTP status code returned by the request.
      * @return data The response body as raw bytes.
      */
-    function get(string memory url) internal returns (uint256 status, bytes memory data) {
+    function get(string memory url) internal returns (uint256, bytes memory) {
+        if (_isPowerShellAvailable())  return _powershell(url);
+        return _bash(url);
```
acuarica, line 34: instead of checking whether powershell is available first, why don't we first try `bash` and if `revert`s then we can try powershell? This would lead to less code (no more `_isPowerShellAvailable`). But most important is that we avoid the extra `ffi` for checking powershell.

README.md
```diff
@@ -57,13 +57,16 @@ ffi = true
...
-> The Foundry library assumes it is running on a Unix-like operating system.
-> In addition to `curl`, it uses `bash`, `tail`, `sed` and `tr` and to make requests and parse responses.
+- On Unix-like operating systems, the library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network. This allows the library to fetch token state from the remote network. Since `curl` is an external command, `ffi` must be enabled.
+- On Windows, if PowerShell is available, the library uses the `Invoke-WebRequest` command to handle HTTP requests, which also requires `ffi` to be enabled.
+
+> [!NOTE]  
+> The Foundry library assumes it is running on either a Windows system with PowerShell or a Unix-like operating system.
+>
+> - On Unix-like systems, it uses `curl`, `bash`, `tail`, `sed`, and `tr` to make requests and parse responses.
+> - On Windows, it uses `Invoke-WebRequest` and `Start-Process` in PowerShell for similar operations.
```
acuarica, line 69:
```suggestion
> - On Windows, it uses `Invoke-WebRequest` and `Start-Process` in PowerShell for a similar effect.
```

README.md
```diff
@@ -57,13 +57,16 @@ ffi = true
 
 Alternatively, you can add the `--ffi` flag to your execution script.
 
-This is necessary because our library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network.
-This enables the library to fetch token state in the remote network.
-Given `curl` is an external command, `ffi` needs to be enabled.
+This is necessary because:
```
acuarica, line 60:
```suggestion
This is necessary because
```

README.md
```diff
@@ -57,13 +57,16 @@ ffi = true
...
 
-> [!NOTE]
-> The Foundry library assumes it is running on a Unix-like operating system.
-> In addition to `curl`, it uses `bash`, `tail`, `sed` and `tr` and to make requests and parse responses.
+- On Unix-like operating systems, the library relies on [`curl`](https://curl.se/) to make HTTP requests to the Hedera remote network. This allows the library to fetch token state from the remote network. Since `curl` is an external command, `ffi` must be enabled.
+- On Windows, if PowerShell is available, the library uses the `Invoke-WebRequest` command to handle HTTP requests, which also requires `ffi` to be enabled.
```
acuarica, line 63: Nice catch!

acuarica (changes requested):

acuarica:
> Perhaps it would be helpful to mention Windows issues with symlinks in the README

what would be the issue? did you have any problems with symlinks on Windows?

In any case, I would say no, because that's for developers of the packages. It won't be a problem for users.

acuarica:
We should also keep an eye on https://github.com/foundry-rs/foundry/issues/5643. If it gets implemented, we can drop bash/powershell altogether.

acuarica (approved):
Nice work!

PR #252, opened 2025-02-27
chore: hardhat example tests in pipeline

### LINKED ISSUE #231: [Hardhat] Run example tests for Hardhat in CI
We should run the example tests under `examples/hardhat-hts` in CI. Add a new job similar to Foundry's example tests `foundry-example-tests`.

arianejasuwienas:
**Description**:

Tests for example hardhat script added.

**Related issue(s)**:

Fixes #231

**Notes for reviewer**:

**Checklist**


.github/workflows/test.yml
```diff
@@ -232,3 +232,44 @@ jobs:
...
+        with:
+          submodules: recursive
+
+      - name: Set up Node.js
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
+        with:
+          node-version: '20'
+
+      - name: Run Forge example tests using remote network with a pinned block number
+        run: cd examples/hardhat-hts && npm install
```
acuarica, line 256: let's use `working-directory:` instead of `cd &&`

.github/workflows/test.yml
```diff
@@ -232,3 +232,44 @@ jobs:
...
+      - name: Making sure the contracts are up to date
+        run: |
+          rm -rf examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+          mv contracts examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+
+      - name: Moving build output to the plugin directory, where it's missing
+        run: mv out examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking
+
+      - name: Run Example tests
+        run: cd examples/hardhat-hts && npx hardhat test --grep USDC && npx hardhat test --grep NFT
```
acuarica, line 275: should we test with no filtering?

acuarica, line 275: there is an npm script to run hardhat tests, `npm run test`, let's use that one instead

.github/workflows/test.yml
```diff
@@ -232,3 +232,44 @@ jobs:
...
+
+      - name: Making sure the contracts are up to date
+        run: |
+          rm -rf examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+          mv contracts examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+
+      - name: Moving build output to the plugin directory, where it's missing
+        run: mv out examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking
+
+      - name: Run Example tests
```
acuarica, line 274:
```suggestion
      - name: Run Hardhat example tests
```

.github/workflows/test.yml
```diff
@@ -232,3 +232,44 @@ jobs:
...
+      - name: Run Forge build
+        run: forge build
+
+      - name: Making sure the contracts are up to date
+        run: |
+          rm -rf examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+          mv contracts examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking/contracts
+
+      - name: Moving build output to the plugin directory, where it's missing
+        run: mv out examples/hardhat-hts/node_modules/@hashgraph/system-contracts-forking
```
acuarica, line 272: let's remove this for now.

.github/workflows/test.yml
```diff
@@ -232,3 +232,44 @@ jobs:
...
+        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+        with:
+          submodules: recursive
+
+      - name: Set up Node.js
+        uses: actions/setup-node@1e60f620b9541d16bece96c5465dc8ee9832be0b # v4.0.3
+        with:
+          node-version: '20'
+
+      - name: Run Forge example tests using remote network with a pinned block number
```
acuarica, line 255: use better name (still using Forge from another job)

acuarica, line 255: move this right before the `npm run test` step

acuarica (changes requested):

.github/workflows/test.yml
arianejasuwienas: @acuarica we can't run the test for creating a new token since it assumes a local node is running, whereas the current approach runs these tests against the mainnet.

.gitignore
```diff
@@ -2,7 +2,10 @@
 
 # Foundry's cache and compiler output files
 /cache/
-/out/
+/out/*
+
+# HTS compiler output needed for `npm` package when consumed directly from GitHub
+!/out/HtsSystemContract.sol/
```
acuarica, line 8: we already went down this path, we should revert this

examples/hardhat-hts/package-lock.json
```diff
@@ -871,7 +871,7 @@
         },
         "node_modules/@hashgraph/system-contracts-forking": {
             "version": "0.1.1",
-            "resolved": "git+ssh://git@github.com/hashgraph/hedera-forking.git#7c592e721c55d29d23052c49cc95db583129f923",
+            "resolved": "git+ssh://git@github.com/hashgraph/hedera-forking.git#c820f975db64e2101d7cde8e7a30212bb6f32ce0",
```
acuarica, line 874: I see the problem, we can consume the forking package in `package.json` in two ways, from npm or directly from local path `@hashgraph/system-contracts-forking": "../.."`. Which one do you think is better?

arianejasuwienas: I went with the '../..' approach because:

1. To use an npm package, we would need to publish it first, which I can't do on my own.
2. The current solution will be more stable, the library will always stay up to date with our repositorys codebase.

acuarica:
@arianejasuwienas npm package published to unblock this due to create token functionality.

acuarica (approved):

arianejasuwienas:
@acuarica LG, thank you!

andrewb1269hg (approved):

PR #259, opened 2025-03-05
chore: switch license declarations to spdx

### LINKED ISSUE #245: Switch license declarations to SPDX
### Problem

Change all license declarations to SPDX format. These should apply only to `.js` files (`.sol` files already have a check for this kind of license). All `.js` files should now start with the following license instead

```
// SPDX-License-Identifier: Apache-2.0
```

> See similar issue in the Relay https://github.com/hashgraph/hedera-json-rpc-relay/issues/3480.

### Solution

Given the new license header is a one-liner, we should inline it in the eslint config and remove the old header `./resources/license.js.header`.

But we need to change or update the package, because the package we are currently using does not support this kind of license. See issue https://github.com/nikku/eslint-plugin-license-header/issues/19#issuecomment-2669602495.

arianejasuwienas:
**Description**:

Switching license declarations to spdx

**Related issue(s)**:

Fixes #245

**Notes for reviewer**:

**Checklist**


acuarica (approved):
lgtm

PR #271, opened 2025-03-17
chore: fix userdoc/devdoc lookup bug

barajeel:
**Description**:

`userdoc[prop]` and `devdoc[prop]` were wrong—`prop` is `"methods"` or `"events"`, but the actual keys are sighashes. This led to missing docs.  

#### **Fix:**  
Pass `userdoc` and `devdoc` directly instead.  

**Before:**  
```js
const members = [...data(kind.toLowerCase(), abi, userdoc[prop] ?? {}, devdoc[prop] ?? {})];
```  

**After:**  
```js
const members = [...data(kind.toLowerCase(), abi, userdoc ?? {}, devdoc ?? {})];
```
**Checklist**

### DIFF
--- a/scripts/abi-table.js
+++ b/scripts/abi-table.js
@@ -45,7 +45,7 @@ function main() {
         ['Function', 'methods'],
         ['Event', 'events'],
     ].forEach(([kind, prop]) => {
-        const members = [...data(kind.toLowerCase(), abi, userdoc[prop] ?? {}, devdoc[prop] ?? {})];
+        const members = [...data(kind.toLowerCase(), abi, userdoc ?? {}, devdoc ?? {})];
         if (members.length > 0) {
             write();
             write(`| ${kind} | Comment |`);


acuarica:
Hi @barajeel, thanks for sending this. The keys for both `userdoc` and `devdoc` are `"methods"` and `"events"`. See for example excerpt from output file `out/IHRC719.sol/IHRC719.json`

```json
    "devdoc": {
        "version": 1,
        "kind": "dev",
        "methods": {
            "associate()": {
                "details": "This function allows an account to opt-in to receive the token",
                "returns": {
                    "responseCode": "The response code indicating the result of the operation"
                }
            },
            "dissociate()": {
                "details": "This function allows an account to opt-out from receiving the token",
                "returns": {
                    "responseCode": "The response code indicating the result of the operation"
                }
            },
            "isAssociated()": {
                "details": "This function returns the association status of the calling account",
                "returns": {
                    "associated": "True if the account is associated, false otherwise"
                }
            }
        }
    },
```

When applying your modification, the output tables does not contain any comment.

I'm closing this as I believe it is not an actual bug. Feel free to reopen if you still think there is a problem here.

PR #281, opened 2025-03-29
docs: improve Foundry HTS example

### LINKED ISSUE #276: [Foundry] Add docs on how to run scripts with `forge`
In addition to https://github.com/hashgraph/hedera-forking/issues/265, we should include in the `README` instructions on how to run Foundry scripts when using HTS tokens.

acuarica:
**Description**:

This PR refactors the Foundry HTS example by creating a proper Foundry project. This simplifies its execution. Moreover, it also includes a Foundry script that showcases how to use our Foundry library to send transactions to a remote network.

**Related issue(s)**:

Fixes #276.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -394,6 +394,13 @@ jobs:
         run: forge script scripts/CreateToken.s.sol -vvv --rpc-url localnode --skip-simulation --broadcast
         working-directory: examples/foundry-hts
 
+      - name: Verify the token was created
+        run: |
+          TOKEN_ADDR=`cat broadcast/CreateToken.s.sol/296/run-latest.json | jq '.receipts[0].contractAddress' --raw-output`
+          TOKEN_CODE=`cast code --rpc-url localnode $TOKEN_ADDR`
+          test $TOKEN_CODE != "0x"
+        working-directory: examples/foundry-hts
+
       - name: Stop the local node
         if: ${{ !cancelled() }}
         run: npx hedera stop
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -348,6 +348,52 @@ jobs:
         run: npm run test
         working-directory: examples/hardhat-hts
 
+  foundry-example-scripts:
+    name: Foundry example scripts w/local node
+    runs-on: smart-contracts-linux-medium
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
+        with:
+          egress-policy: audit
+
+      - name: Checkout Code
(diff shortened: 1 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -349,6 +348,68 @@ jobs:
...
+        timeout-minutes: 5
+
+      - name: Run Forge build on Foundry HTS example
+        run: forge build
+        working-directory: examples/foundry-hts
+
+      # Uses a local node's account private key
+      # https://github.com/hiero-ledger/hiero-local-node?tab=readme-ov-file#hedera-start-options
+      - name: Set up Local Node's account Private Key
+        run: echo "PRIVATE_KEY=0x105d050185ccb907fba04dd92d8de9e32c18305e097ab41dadda21489a211524" > .env
```
quiet-node, line 392: Ik this is a private key of a local-node account but seeing a key hard-coded makes me a little angsty. Is it possible to leverage the .env in the /example/foundry-hts. Just thinking out loud no big deal.

quiet-node (approved):
LGTM just two thoughts

.github/workflows/test.yml
```diff
@@ -349,6 +348,68 @@ jobs:
...
+      - name: Install Dependencies
+        run: npm ci
+
+      - name: Start the local node
+        run: npx hedera start --detach --verbose=trace
+        timeout-minutes: 5
+
+      - name: Run Forge build on Foundry HTS example
+        run: forge build
+        working-directory: examples/foundry-hts
```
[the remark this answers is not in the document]
acuarica: Sure, so the first one runs on the root of the repo. This one runs on the `examples/foundry-hts` folder. This folder contains the Foundry example that uses the library located in the root folder.

.github/workflows/test.yml
acuarica: As Logan mentioned offline, the Relay uses .env.example with those private key(s). The issue is that I wouldn't want to pollute the example, and keep it simple as possible. The user most probably would use a previewnet/testnet PK, so including a local node one would only be used here.

rbarker-dev: Best practice would have us add this as a GitHub secret or environment variable. I would recommend using the .env.example file here.

rbarker-dev (commented):
One nit.

.github/workflows/test.yml
acuarica: > I would recommend using the .env.example file here.

Thanks Roger for taking a look. As mentioned above https://github.com/hashgraph/hedera-forking/pull/281#discussion_r2067441865, using a local node's PK is not really an example. It's only for testing, and for this testing. The only reason the PK is not embedded in the `.s.sol` file is because this way it can be reused to point to a different network. Also it serves as documentation for the `README.md`.

For reference take a look at `test/hts.e2e.js` where a couple of PKs are embedded there because that's only used for testing purposes.

PR #282, opened 2025-04-01
fix: incorrect usage of `sighash` for events

barajeel:
**Description**:

`sighash` was being applied to events, which isn’t correct—events use `topicHash` instead.
This update ensures that functions use `sighash`, while events correctly use `topicHash`.

**Checklist**

### DIFF
--- a/scripts/abi-table.js
+++ b/scripts/abi-table.js
@@ -15,7 +15,7 @@ function* data(kind, abi, userdoc, devdoc) {
     for (const entry of abi) {
         const frag = Fragment.from(entry);
         if (frag.type === kind) {
-            const sighash = frag.format('sighash');
+            const sighash = kind === 'function' ? frag.format('sighash') : frag.format('topicHash');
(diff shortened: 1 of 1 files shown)


acuarica:
Hi @barajeel, thanks for putting the effort in sending this. I've tried your PR locally but it fails to get the documentation for events. For example if I run `./scripts/abi-table.js out/IERC20.sol/IERC20.json` I get

```markdown
| Event | Comment |
|----------|---------|
| `Approval(address indexed owner, address indexed spender, uint256 amount)` |   |
| `Transfer(address indexed from, address indexed to, uint256 amount)` |   |
```

Please note that `topicHash` is not a valid value for the `format` method. See Ethers' [FormatType](https://github.com/ethers-io/ethers.js/blob/0195f440e35d11f84874f58d80b2071f52c4b17f/src.ts/abi/fragments.ts#L115) definition.

I'm going to close this. Feel free to re-open if you see any other problem.

PR #291, opened 2025-05-11
feat: add HTS foundry scripts

### LINKED ISSUE #292: [Foundry] Make `associate`/`disassociate` not revert when account does not exist
Methods `associate`, `disassociate` and `isAssociated` use the same slot to determine the token association. In particular, the `isAssociated` method `revert`s when the provided address does not belong to an existing Hedera account. However, both `associate` and `disassociate` should not revert in this situation.

Improve the association methods so they don't revert when the provided address does not belong to a Hedera account.

acuarica:
**Description**:

This PR fixes the association methods to not `revert` when the provided address does not belong to an existing Hedera account. Scripts are included so the behavior can be validated against a real network.

> [!IMPORTANT]
> Contracts are also added into the `script` folder, because haveing a `src` folder in the example causes the remappings in the user's project to fail.

**Related issue(s)**:

Fixes #292.

**Notes for reviewer**:

**Checklist**


quiet-node (approved):
LGTM

